PR #1, opened 2025-10-03
feature/initialize-a-react-project-with-vite

sabellius:
## Description
This branch initializes a React project using Vite just as we did in class. I also added some formatting rules for EditorConfig and Prettier.js with the rules we agreed on in today's meeting. 

For the formatting rules to apply in your editor (assuming we all use vscode), you need to install the Prettier and EditorConfig extensions from the marketplace.

Please let me know if you have any questions.

## Changes
- Initialize a React project using Vite with the following the guidelines we got in Class (framework: React,  variant: Javascript, Experimental: false).
- Add an EditorConfig configuration file with same sane defaults (most conform with Prettier's defaults).
For more information, see: [EditorConfig: Example file](https://editorconfig.org/#example-file)
- Add a Prettier.js configuration file with the default options we agreed on in today's meeting, with the exception of trailing commas, and arrow function parentheses. For more information, see: [Pettier: Options](https://prettier.io/docs/options)


LiorRey:
Sounds good to me, merged 👍

PR #3, opened 2025-10-12
style: Adjusted layout CSS for board page

snoxlax:
#### What I did

I created a board component with the following structure:

- **BoardDetails** - Main board component that displays lists
- **List** - Individual list component that contains cards
- **Card** - Individual card component within each list

- **BoardIndex.jsx** - Contains template data and renders one BoardDetails component
- **BoardDetails.jsx** - Represents a single board with lists and cards
- **List.jsx** - Displays cards within a list
- **Card.jsx** - Individual card items

##### Data Structure

```javascript
// Template data in BoardIndex.jsx
const boardObject = {
  id: "board1",
  title: "Template Board",
  lists: [
    {
      id: "list1",
      title: "Template List 1",
      cards: [
        { id: "card1", title: "Card 1", description: "..." },
        { id: "card2", title: "Card 2", description: "..." },
      ],
    },
    // ... more lists
  ],
};
```

##### Styling

- **CSS Grid** for main layout structure
- **Horizontal scrolling** for lists when they exceed container width
- **Fixed-width lists** (350px) with proper spacing
- **Full-height layout** using `grid-template-rows: auto 1fr auto`

##### Features

- Board displays multiple lists horizontally
- Each list contains multiple cards
- Horizontal scroll when lists don't fit in container
- changed the main layout a bit to be more simple


LiorRey:
I think the name **List** is already taken for a list of board/cards/etc.
If we want to display a list of boards or a list of cards, we’d use BoardList and CardList respectively, so if we go with **List** instead of **Group**, we’d end up with "ListList.jsx".
So I think we should choose the term **Group**.

But the term you suggested, **Card** - is great and fits better than **Task** in my opinion, because the items inside a **Group** aren’t necessarily tasks. A card might contain a general note or a contact’s email and phone number, not necessarily a to-do item. So the more general term, **Card**, feels more appropriate to me than **Task**.

In short, I think the entities should be:

- **Board**
- **Group** (instead of List)
- **Card** (instead of Task)

But nonetheless, kudos to you! Awesome job! 🦾

snoxlax:
I continued working on the layout, styling, buttons, and navigation bars.

#### Board object example (for now)
```8:40:src/services/car/car-service-local.js
id: "board-1",
name: "Project Alpha",
description: "Main development board for Project Alpha",
createdAt: "2025-10-12T10:00:00Z",
updatedAt: "2025-10-12T16:00:00Z",
lists: [
  {
    id: "list-1",
    name: "To Do",
    position: 1,
    cards: [
      {
        id: "card-12",
        title: "Set up project repo",
        description: "Initialize repository and push base structure",
        labels: [
          { name: "frontend", color: "blue" },
          { name: "priority:high", color: "green" },
        ],
        assignedTo: ["user-1"],
        createdAt: "2025-10-12T10:05:00Z",
        dueDate: "2025-10-15T00:00:00Z",
      },
      {
        id: "card-13",
        title: "Define API endpoints",
        description: "Draft a list of REST endpoints for backend",
        labels: [{ name: "frontend", color: "blue" }],
        assignedTo: ["user-2"],
        createdAt: "2025-10-12T11:00:00Z",
      },
    ],
  },
// ... more code ...
```

#### What changed
- **Layout**: Continued refining structure to better present lists and cards.
- **Styling**: Iterated on base/component styles for consistency.
- **Buttons**: Created icon style buttons in specials.css with Lucide React.
- **Nav bars**: Improved header/footer navigation and spacing.
- and more...

#### Notes
- Frontend-only updates; leverages the board data model already in `src/services/car/car-service-local.js`.

snoxlax:
@LiorRey 
> * **Board**
> * **Group** (instead of List)
> * **Card** (instead of Task)

In my opinion, we can go with just List, Board, and Card. I don't think we will have a ListList because right now, within a single board, I map over a collection of lists, and each iteration renders a List component.

So I think we can go with just the name and then if needed simply add the plural form like Lists, Boards, Cards.

package.json
```diff
@@ -10,11 +10,12 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "@hookform/resolvers": "^5.2.2",
     "axios": "^1.12.2",
+    "lucide-react": "^0.545.0",
```
sabellius, line 18: I'm not familiar with this icon library, but material ui also offers an icon library. What made you decide choosing this library over more popular ones? Is this one well maintained? Is it secure?

src/assets/styles/basics/specials.css
```diff
@@ -0,0 +1,19 @@
+/* dots button */
```
sabellius, line 1: What is this file? Couldn't quit understand just from the name. What does this contain?

src/assets/styles/main.css
```diff
@@ -20,11 +20,10 @@
 @import "components/car/CarList";
```
sabellius, line 20: Since you've renamed the components and essentially removed them, you also need to rename/remove those imports.

src/components/Card.jsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
```
sabellius: I don't think you need this line since some version of react. This comment applies to all places where it is written.

src/components/Card.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default function Card({ card, onRemoveCard, onUpdateCard }) {
```
sabellius, line 1: We need to agree whether we `export` or `export default` components for code consistency.

src/components/Card.jsx
```diff
@@ -0,0 +1,31 @@
...
+  function onRemoveCard() {
+    onRemoveCard(card.id);
+  }
+
+  function onUpdateCard() {
```
sabellius, line 6: The above functions are defined but not used in the component, is this by mistake?

src/components/Footer.jsx
```diff
@@ -1,18 +1,44 @@
...
+        <FooterButton>
+          <SquareKanban />
+          Switch Board
+        </FooterButton>
+      </div>
     </footer>
   );
 }
+
+function FooterButton(props) {
```
sabellius, line 29: In my opinion, a `.jsx` file should contain only one component. What is the purpose of this component? how is it different from a regular button or a link?

src/components/List.jsx
```diff
@@ -0,0 +1,46 @@
+import React, { useState, useEffect } from "react";
+import Card from "./Card";
+import { Ellipsis } from "lucide-react";
+export default function List({ list, onRemoveList, onUpdateList }) {
```
sabellius, line 5: Formatting here is a bit off. Please add some blank lines between the imports and the components, and between functions.

src/components/List.jsx
```diff
@@ -0,0 +1,46 @@
...
+  useEffect(() => {
+    setCards(list.cards);
+  }, [list.cards]);
+
+  async function onAddCard() {}
+  async function onRemoveCard(cardId) {
+    await onRemoveList(list.id);
+  }
+
+  async function onUpdateCard(card) {
```
sabellius, line 18: I'm not really sure if the component knows whether it receives a sync or async function as props. In my opinion the "React way" to do what you did here is an inline anonymous function:
```
            <Card
              key={card.id}
              card={card}
              onRemoveCard={() => onRemoveCard(card.id)}
              onUpdateCard={() => onUpdateCard(card)}
            />
```

src/pages/BoardDetails.jsx
```diff
@@ -0,0 +1,61 @@
...
+        </ul>
+        <nav className="board-footer">
+          <Footer />
+        </nav>
+      </div>
+    </section>
+  );
+}
+
+function AddListButton() {
```
sabellius, line 55: Same here, In my opinion there is not reason to separate this into a different component, why not just use the html in the first component?

src/services/car/car-service-local.js
```diff
@@ -4,15 +4,287 @@ import { userService } from "../user";
 
 const STORAGE_KEY = "car";
 
+const boardObject = {
```
sabellius, line 7: Is this the structure we're going with? How did you generate it?

package.json
snoxlax: We could use MaterialUI i just installed this beacsue i needed just some simple icons and not the whole MUI framework
Also we decided to use MUI in the end? or maybe some other framework?
but we can use MUI too they have the same icons

lucide-react is just for icons and its open source and is actively maintained by the Lucide community and contributors on GitHub https://lucide.dev/

src/assets/styles/basics/specials.css
snoxlax: This is a seprate css file I'm not sure about the name yet could be more suitable to call it custom-styles?
its soppsued to be a reusable CSS for now it contains the style of the icon button beacuse its been used in many places
its reusable style componnets

src/assets/styles/main.css
snoxlax: No they are still there I made new css files instead if we agree we dont need these we can delete them 
and I didnt touch UserMessage almost all the starter css files are still there

src/components/Card.jsx
snoxlax: I’m used to using export default, but I know they taught us not to use it. I’m not entirely sure why, though.

src/components/Card.jsx
snoxlax: Its just a placeholder they dont do anything its also the same in List

src/components/Footer.jsx
snoxlax: I’ve mentioned this a few times before. This particular Button could be more global, but since it’s only used inside the footer for now, I think we should keep it as a local component specific to the footer it’s kind of unique to this area and isn’t reusable elsewhere for now.

But we can decide together whether we want to follow a rule of one component per file.

src/pages/BoardDetails.jsx
snoxlax: You are right about this one we could seprate it to a diffrent componnet

src/services/car/car-service-local.js
snoxlax: This is from ChatGpt its just for now I also edited it to have label { title:something, color:red }
Its just to display a board

src/components/List.jsx
snoxlax: Its the same as before its just a place holder and it was from the starter this way

LiorRey (approved):
We've discussed on Discord and settled the things 👍

snoxlax:
Ive now also merged the main with the MUI package and the new .gitattributes file in this pr 
so just waiting after @LiorRey 's pr merge and @sabellius approval

LiorRey:
Merged my PR (https://github.com/LiorRey/ca-final-project/pull/4) 👍

PR #4, opened 2025-10-13
feat: Set agreed routes structure with required fixes

LiorRey:
**What I did :**

In `App.jsx`, I set the routes structure as confirmed by Tal :

```
import { Routes, Route } from "react-router";
import { Provider } from "react-redux";

import { store } from "./store/store.js";
import { AppHeader } from "./components/AppHeader.jsx";
import { UserMessage } from "./components/UserMessage.jsx";
import { HomePage } from "./pages/HomePage.jsx";
import { AboutUs, AboutTeam, AboutVision } from "./pages/AboutUs.jsx";
import { BoardIndex } from "./pages/BoardIndex.jsx";
import { BoardDetails } from "./pages/BoardDetails.jsx";
import { CardDetails } from "./pages/CardDetails.jsx";
import { UserDetails } from "./pages/UserDetails.jsx";
import { ChatApp } from "./pages/Chat.jsx";
import { AdminIndex } from "./pages/AdminIndex.jsx";
import { LoginSignup, Login, Signup } from "./pages/LoginSignup.jsx";
import { AppFooter } from "./components/AppFooter.jsx";

export function App() {
  return (
    <Provider store={store}>
      <div className="main-container">
        <AppHeader />
        <UserMessage />

        <main>
          <Routes>
            <Route path="" element={<HomePage />} />
            <Route path="about" element={<AboutUs />}>
              <Route path="team" element={<AboutTeam />} />
              <Route path="vision" element={<AboutVision />} />
            </Route>
            <Route path="board" element={<BoardIndex />} />
            <Route path="board/:boardId" element={<BoardDetails />} />
            <Route
              path="board/:boardId/:groupId/:cardId"
              element={<CardDetails />}
            />
            <Route path="user/:userId" element={<UserDetails />} />
            <Route path="chat" element={<ChatApp />} />
            <Route path="admin" element={<AdminIndex />} />
            <Route path="auth" element={<LoginSignup />}>
              <Route path="login" element={<Login />} />
              <Route path="signup" element={<Signup />} />
            </Route>
          </Routes>
        </main>
        <AppFooter />
      </div>
    </Provider>
  );
}

```

I added the following initial JSX files for their respective routable components under `pages` folder :

- `BoardIndex.jsx`
- `BoardDetails.jsx`
- `CardDetails.jsx`

I also removed the files `Header.jsx` and `Footer.jsx`, in favor of adding the better-named files `AppHeader.jsx` and `AppFooter.jsx`, respectively (each has the same code that was in the previous file, but considering the new file name).

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,8 @@ src/
 
 ### Core Components
 
-- `AppHeader` - Navigation and user menu
-- `AppFooter` - Footer with service status
+- `Header` - Navigation and user menu
+- `Footer` - Footer with service status
 - `UserMsg` - Toast notifications
 - `CarList` - Grid display of cars with actions
 - `CarFilter` - Search and filter interface
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,47 +1,51 @@
-import React from "react";
 import { Routes, Route } from "react-router";
+import { Provider } from "react-redux";
 
-import { HomePage } from "./pages/HomePage";
-import { AboutUs, AboutTeam, AboutVision } from "./pages/AboutUs";
-import { CarIndex } from "./pages/CarIndex.jsx";
-import { ReviewIndex } from "./pages/ReviewIndex.jsx";
+import { store } from "./store/store.js";
+import { AppHeader } from "./components/AppHeader.jsx";
+import { UserMessage } from "./components/UserMessage.jsx";
+import { HomePage } from "./pages/HomePage.jsx";
+import { AboutUs, AboutTeam, AboutVision } from "./pages/AboutUs.jsx";
+import { BoardIndex } from "./pages/BoardIndex.jsx";
+import { BoardDetails } from "./pages/BoardDetails.jsx";
+import { CardDetails } from "./pages/CardDetails.jsx";
+import { UserDetails } from "./pages/UserDetails.jsx";
 import { ChatApp } from "./pages/Chat.jsx";
 import { AdminIndex } from "./pages/AdminIndex.jsx";
-
-import { CarDetails } from "./pages/CarDetails";
-import { UserDetails } from "./pages/UserDetails";
-
-import { Header } from "./components/Header";
-import { Footer } from "./components/Footer";
-import { UserMessage } from "./components/UserMessage.jsx";
 import { LoginSignup, Login, Signup } from "./pages/LoginSignup.jsx";
+import { AppFooter } from "./components/AppFooter.jsx";
 
 export function App() {
   return (
-    <div className="main-container">
-      <Header />
-      <UserMessage />
+    <Provider store={store}>
+      <div className="main-container">
+        <AppHeader />
+        <UserMessage />
 
-      <main>
-        <Routes>
-          <Route path="" element={<HomePage />} />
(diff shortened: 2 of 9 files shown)


snoxlax:
Some things here override what we agreed from before 
Did you work on the up-to-date version of main? we agreed that AppHeader would be Header and AppFooter would be just Footer for example

sabellius:
I'm a little bit confused as there are new imports but no changed files to back those imports. For example, `AboutUs` page, but I can't find the components imported in this PR, so you might have not pushed all the relevant files yet. Regardless, In my opinion (and it's most likely a React standard coding style), that a `.jsx` file should contain only one component. So I would either split it into separate files.

Another question about this page, is I see that `AboutTeam`. and `AboutVision` are nested routes under `About`. Is that necessary? Wouldn't it be simpler for the about page to contain all those components inside it or perhaps we display it differently in the UI and I misunderstood something.
@LiorRey @snoxlax

snoxlax:
I think it’s just something that was in the starter project, and I agree it shouldn’t be nested. (`About`, `AboutUs`, `AboutTeam`)
he just removed it and added it back in the imports

I also agree that we shouldn’t have multiple exports from a single component, but I think it’s fine to have some smaller components inside the main component. For example, if we have Card like this…


#### Example
```jsx
// Card.jsx
export default function Card({ title, children }) {
  return (
    <article className="card">
      <Lable title={title} />
      ...
    </article>
  );
}

function Lable({ title }) {
  return <span className="card-lable">{title}</h3>;
}
```

I think it’s better this way, especially since we don’t need to reuse the Label component elsewhere. Keeping it local means fewer files and simpler structure. The labels are only used within this component, and since we have multiple labels inside the Card, it makes sense to create a local component specifically for the Card component.

what do you think?

LiorRey:
Fixes :

- Back to `Header.jsx` and `Footer.jsx` (no App prefix)

- Using `listId` (instead of `groupId`) in :
`<Route path="board/:boardId/:listId/:cardId" element={<CardDetails />} />`

snoxlax (approved):
Looks good to me!

sabellius:
> I think it’s just something that was in the starter project, and I agree it shouldn’t be nested. (`About`, `AboutUs`, `AboutTeam`)
> he just removed it and added it back in the imports
> 
> I also agree that we shouldn’t have multiple exports from a single component, but I think it’s fine to have some smaller components inside the main component. For example, if we have Card like this…
> 
> 
> #### Example
> ```jsx
> // Card.jsx
> export default function Card({ title, children }) {
>   return (
>     <article className="card">
>       <Lable title={title} />
>       ...
>     </article>
>   );
> }
> 
> function Lable({ title }) {
>   return <span className="card-lable">{title}</h3>;
> }
> ```
> 
> I think it’s better this way, especially since we don’t need to reuse the Label component elsewhere. Keeping it local means fewer files and simpler structure. The labels are only used within this component, and since we have multiple labels inside the Card, it makes sense to create a local component specifically for the Card component.
> 
> what do you think?

I understand and agree with the idea behind this, just not used to doing this with components xd. Looks good to me :)

sabellius (approved):

PR #5, opened 2025-10-13
feature/CATC_16-add-board-state-to-a-global-state

sabellius:
## Description
This branch adds a reducer and actions for managing board state globally across the app.

## Changes
- Add a board reducer that also includes loading and error states.
- Add actions for changing board state.

## Notes
- I used a common pattern than we haven't used in class before but can be found in the starter code, which is action creators - pure functions that return an action object. This decouples components from action structure, and allows for easier refactoring of the action structure when needed (instead of updating object literals in multiple places).
- I also the same key for all actions updated data - `payload`. In my opinion this makes it more intuitive to add additional cases in the reducer, and generally easier to work with actions as you know what structure to anticipate.

### DIFF
--- a/src/store/actions/board-actions.js
+++ b/src/store/actions/board-actions.js
@@ -2,7 +2,7 @@ import {
   SET_BOARDS,
   ADD_BOARD,
   UPDATE_BOARD,
-  REMOVE_BOARD,
+  DELETE_BOARD,
   SET_BOARD,
   SET_LOADING,
   SET_ERROR,
@@ -37,10 +37,7 @@ export async function loadBoard(boardId) {
 
 export async function createBoard(board) {
   try {
-    const newBoard = await boardService.updateBoardWithActivity(board, {
-      key: "created board",
-      value: "",
-    });
+    const newBoard = await boardService.save(board);
     store.dispatch(addBoard(newBoard));
     return newBoard;
   } catch (error) {
@@ -49,9 +46,17 @@ export async function createBoard(board) {
   }
 }
 
-export async function updateBoard(board) {
+export async function updateBoard(
+  board,
+  { listId = null, cardId = null, key, value }
+) {
   try {
-    const updatedBoard = await boardService.updateBoardWithActivity(board);
+    const updatedBoard = await boardService.updateBoardWithActivity(board, {
+      listId,
+      cardId,
+      key,
+      value,
+    });
     store.dispatch(editBoard(updatedBoard));
     return updatedBoard;
   } catch (error) {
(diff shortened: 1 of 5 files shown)


src/store/store.js
snoxlax, line 1: Not really a big deal but shouldnt it be `boardsModule` or others changed to just `user` `system` etc. which could a bit vague?

snoxlax (approved):
Looks good to me!

src/store/store.js
sabellius: I didn't use `boardsMdoule` but rather, `boards`. I kept `usersModule` because it's used somewhere in the header component and throws errors if I comment/remove it. I would usually create a new ticket for renaming this, but there is already a similar ticket to this pull request, to add user state to the global store, which is already implemented in the starter in one way or another. So I can rename it as part of that ticket, while refactoring what needs to be changed there (if any).

LiorRey: I liked the use of `payload` and the concept of createAction.
As snoxlax said, it is not a big deal but why not using `boardModule` ?
But I approve this PR nonetheless 🦾

LiorRey (approved):

sabellius:
@LiorRey @snoxlax I'll merge this one once [Create a service for boards](https://trello.com/c/48hIbR8W) is merged.

src/store/actions/board-actions.js
```diff
@@ -0,0 +1,99 @@
...
+export async function updateBoard(board) {
+  try {
+    const updatedBoard = await boardService.updateBoardWithActivity(board);
+    store.dispatch(editBoard(updatedBoard));
+    return updatedBoard;
+  } catch (error) {
+    store.dispatch(setError(`Error updating board: ${error.message}`));
+    throw error;
+  }
+}
```
sabellius, line 66: @LiorRey @snoxlax It work but its kinda problematic using `updateBoardWithActivity` in the action. I think that whoever wants to dispatch this action should call `updateBoardWithActivity` and then, call the `updateBoard` or `createBoard` function with the updated object. Then, those two function will only have to call `boardService.save()` with the updated object just like we always did so far. What you guys think?

snoxlax: I dont fully understand what you mean do you mean the frontend should call the service of `updateBoardWithActivity` and than call the action? so it would be...

Front --> service.updateBoardWithActivity()
Front --> Action -> updateBoard | createBoard
Action --> service.save();

something like this?

sabellius: Yes, something like this. But after thinking about it, it doesn't makes much sense. Instead I'm using the regular save inside the `createBoard` action (in which we don't need `key`, `value`, `listId`, nor `cardId`). I also modified the `boardService.createBoardWithActivity` function and the `updateBoard` action to use an object as the second argument instead of passing multiple arguments. This will be easier to work with in terms of not relying on argument order, which will allow us to easily add/remove function arguments in the future. It also makes function calls cleaner as they only receive 2 arguments. What do You guys think? @LiorRey @snoxlax

LiorRey: @sabellius Could you please clarify what you meant? I'm not sure I fully understand the suggested flow.
Also, the signature of the `boardService.updateBoardWithActivity()` function is:
`async function updateBoardWithActivity(board, { key, value }, listId = null, cardId = null)`
So in line 54 above, we also need to provide a `{ key, value }` pair and optionally `listId` and `cardId`.

PR #6, opened 2025-10-13
feat: Create a local service for boards

LiorRey:
**Feature:** Local Board Service Implementation
**Branch:** feature/CATC_5-create-local-service-for-boards
**Related Ticket:** CATC-5
___
### 💁‍♂️ What I did
I added an initial local board service enabling CRUD operations on boards, lists, and cards
using local storage via the `async-storage-service`.
It also includes activity tracking for updates and initializes a demo board for testing.
___
### 🗂️ Added Files

**🆕 services/board/board-service-local.js**
Implements all local data logic:
- `query()` - Fetch all boards
- `getById(boardId)` - Get a specific board
- `save(board)` – Create or update a board
- `updateBoard(board, listId, cardId, { key, value })` - Update board/list/card fields and log activity
- `remove(boardId)` - Delete a board
- `createActivity()` - Log board updates
- `_createBoards()` - Load boards from local storage if any exist; otherwise - initialize with sample demo board
- `_createBoard()` - Return the “Project Alpha” demo board

**🆕 services/board/index.js**
- Enables potential switching between local and remote services based on VITE_LOCAL (after remote storage will be added in the future, as we currently have only local storage)
- Provides getEmptyBoard() utility
- Exposes boardService globally in dev mode for easier debugging
___
### ⚙️ Highlights
- Simulates full CRUD functionality without a backend
- Automatically logs activities for updates
- Includes preloaded demo data for immediate visualization
- Async methods mirror future remote service API structure

### DIFF
--- a/src/services/board/board-service-local.js
+++ b/src/services/board/board-service-local.js
@@ -15,60 +15,54 @@ export const boardService = {
 window.cs = boardService;
 
 async function query() {
-  return await storageService.query(BOARDS_STORAGE_KEY);
+  try {
+    return await storageService.query(BOARDS_STORAGE_KEY);
+  } catch (error) {
+    console.log("Cannot load boards:", error);
+    return null;
+  }
 }
 
 function getById(boardId) {
-  let board = null;
-
   try {
-    board = storageService.get(BOARDS_STORAGE_KEY, boardId);
+    return storageService.get(BOARDS_STORAGE_KEY, boardId);
   } catch (error) {
     console.log("Cannot get board:", error);
+    return null;
   }
-
-  return board;
 }
 
 async function save(board) {
-  let savedBoard = null;
-
   try {
-    const boardToSave = { ...board };
-
     if (board._id) {
-      savedBoard = await storageService.put(BOARDS_STORAGE_KEY, boardToSave);
+      return await storageService.put(BOARDS_STORAGE_KEY, { ...board });
     } else {
-      savedBoard = await storageService.post(BOARDS_STORAGE_KEY, boardToSave);
+      return await storageService.post(BOARDS_STORAGE_KEY, { ...board });
     }
-
-    return savedBoard;
   } catch (error) {
     console.log("Cannot save board:", error);
+    return null;
   }
 }
 
 async function updateBoard(board, listId, cardId, { key, value }) {
-  let prevValue = "";
-  let activity = {};
+  let prevValue;
 
   try {
-    if (!board || !key) throw "No board or key provided";
+    if (!board || !key) throw new Error("No board or key provided");
 
-    const listIndex = board.lists.findIndex(list => list.id === listId);
-    if (listIndex !== -1) {
-      const cardIndex = board.lists[listIndex].cards.findIndex(
-        card => card.id === cardId
-      );
+    const list = board.lists?.find(l => l.id === listId);
+    if (list) {
+      const card = list.cards?.find(c => c.id === cardId);
 
-      if (cardIndex !== -1) {
+      if (card) {
         // Update a Card field
-        prevValue = board.lists[listIndex].cards[cardIndex][key];
-        board.lists[listIndex].cards[cardIndex][key] = value;
+        prevValue = card[key];
+        card[key] = value;
       } else {
         // Update a List field
-        prevValue = board.lists[listIndex][key];
-        board.lists[listIndex][key] = value;
+        prevValue = list[key];
+        list[key] = value;
       }
     } else {
       // Update a Board field
@@ -76,10 +70,10 @@ async function updateBoard(board, listId, cardId, { key, value }) {
       board[key] = value;
     }
   } catch (error) {
-    console.log("Cannot save board:", error);
+    console.log("Cannot update board:", error);
   } finally {
     if (board && key) {
-      activity = createActivity(
+      const activity = createActivity(
         board._id,
         listId || null,
         cardId || null,
@@ -92,6 +86,8 @@ async function updateBoard(board, listId, cardId, { key, value }) {
       board.activities.unshift(activity);
     }
   }
+
+  return board;
 }
 
 async function remove(boardId) {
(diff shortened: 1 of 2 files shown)


src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+    const listIndex = board.lists.findIndex(list => list.id === listId);
+    if (listIndex !== -1) {
+      const cardIndex = board.lists[listIndex].cards.findIndex(
+        card => card.id === cardId
+      );
+
+      if (cardIndex !== -1) {
+        // Update a Card field
+        prevValue = board.lists[listIndex].cards[cardIndex][key];
+        board.lists[listIndex].cards[cardIndex][key] = value;
```
sabellius: You can make this a little more DRY by putting the card in a variable. Something like:
```js
const card = board.lists[listIndex].cards[cardIndex];
prevValue = card[key];
card[key] = value
```
The same applies to the other conditional clauses.

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  return board;
+}
+
+async function save(board) {
+  let savedBoard = null;
+
+  try {
+    const boardToSave = { ...board };
```
sabellius: Why do we need a new object here?

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+    board = storageService.get(BOARDS_STORAGE_KEY, boardId);
+  } catch (error) {
+    console.log("Cannot get board:", error);
+  }
+
+  return board;
+}
+
+async function save(board) {
+  let savedBoard = null;
```
sabellius: It doesn't really matter here, but you can declare a variable without giving it an initial value first - `let board;`. The same goes to other places it's done as well, unless there is a purpose for assigning it to `null` first.

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+  getById,
+  save,
+  updateBoard,
+  remove,
+  createActivity,
+};
+window.cs = boardService;
+
+async function query() {
+  return await storageService.query(BOARDS_STORAGE_KEY);
```
sabellius: Missing error handling

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+      savedBoard = await storageService.post(BOARDS_STORAGE_KEY, boardToSave);
+    }
+
+    return savedBoard;
+  } catch (error) {
+    console.log("Cannot save board:", error);
+  }
+}
+
+async function updateBoard(board, listId, cardId, { key, value }) {
```
sabellius: This function is kinda complex, perhaps we could benefit from breaking it down to smaller, easier to understand parts?
maybe something like:
```js
if (listId, && cardId) {
  setCardValue(...)
} else if (listId) {
  setListValue(...)
} else {
  // handle board value
}
```

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+  let prevValue = "";
+  let activity = {};
+
+  try {
+    if (!board || !key) throw "No board or key provided";
+
+    const listIndex = board.lists.findIndex(list => list.id === listId);
```
sabellius: If I'm not mistaken, there is not benefit working with indexes here. Regardless of whether you're working with a card or a list here you can save the list in a variable first, and then use newer js features such as optional chaining to remove some nesting caused by if conditions:
```js
const list = board.lists?.find(list => list.id === listId);
const card = list?.cards?.find(card => card.id === cardId);
```

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+    console.log("Cannot save board:", error);
+  }
+}
+
+async function updateBoard(board, listId, cardId, { key, value }) {
+  let prevValue = "";
+  let activity = {};
+
+  try {
+    if (!board || !key) throw "No board or key provided";
```
sabellius: Why do you perform this check here but not in other places that access or modify boards or their properties?

src/services/board/board-service-local.js
snoxlax: I agree we dont need indexes beacuse we can find by reference with `find()`

snoxlax (approved):
Other than what @sabellius already pointed out looks good to me!

src/services/board/board-service-local.js
snoxlax: I dont know about that in my opinion its not that complex isnt it like 2 lines for each? 😅
also we dont reuse it anywhere else 

maybe if there was one function like this?..

```js
function setObjectValue(object, value){
// update the object (list/card)
}
```

but I dont know looks ok to me

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+  let boards = loadFromStorage(BOARDS_STORAGE_KEY);
+  if (!boards || !boards.length) {
+    boards = [];
+    boards.push(_createBoard());
+
+    saveToStorage(BOARDS_STORAGE_KEY, boards);
+  }
+}
+
+function _createBoard() {
```
snoxlax, line 185: This will create a single board with one list. Shouldn’t we create something more varied like a board with several lists and different card counts to test the board more effectively?

src/services/board/board-service-local.js
sabellius: @snoxlax I would normally agree, but when a function is almost 50 lines long, in many cases it has too much responsibility and complexity. The same goes when there is a comment in each block of code explaining what the code does instead of the code itself being self-explanatory.

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,226 @@
...
+    } else {
+      // Update a Board field
+      prevValue = board[key];
+      board[key] = value;
+    }
+  } catch (error) {
+    console.log("Cannot save board:", error);
+  } finally {
+    if (board && key) {
+      activity = createActivity(
```
sabellius: Should an activity be created even if the update failed?
Also, in terms of SOLID code, should this function really be responsible for creating an activity? I get that it's convenient to put this code here, but then this function not only updates a board object, but also creates an entirely different entity. In my opinion this is a bad practice and although we're trying to build the app fast, this "quick and dirty" methodology will quickly make the code a nightmare to work with.

src/services/board/board-service-local.js
sabellius: @snoxlax There is a [ticket](https://trello.com/c/bVDrQB6s) for this in which functions for creating various entities (board/list/card) with more dynamic content should be implemented. In my opinion this could be used for now until that ticket is finished. What do you think?

src/services/board/board-service-local.js
LiorRey: DRIED it 👍

src/services/board/board-service-local.js
LiorRey: Got rid of those objects 👍

src/services/board/board-service-local.js
LiorRey: Got rid of those objects, returning null in catch-clause only👍

src/services/board/board-service-local.js
LiorRey: Added error handling 👍

src/services/board/board-service-local.js
LiorRey: Tried to simplify `updateBoard` function even more (which in my opinion too was not that complex but sure is a bit long) but as for now it is not in use and I'm not sure if it should be merged with `save` function into one unified function of save/update. I would like to clarify that with Tal in our next meeting.

src/services/board/board-service-local.js
LiorRey: Not using indices anymore, now using optional chaining 👍

src/services/board/board-service-local.js
LiorRey: Added more lists and cards to the demo board 👍

LiorRey:
Thank you for your feedback!
I've updated the code according to it 👍
@sabellius @snoxlax

src/services/board/board-service-local.js
sabellius: @LiorRey ^^^

src/services/board/board-service-local.js
sabellius: @LiorRey ^^^

src/services/board/board-service-local.js
```diff
@@ -0,0 +1,382 @@
...
+      } else {
+        // Update a List field
+        prevValue = list[key];
+        list[key] = value;
+      }
+    } else {
+      // Update a Board field
+      prevValue = board[key];
+      board[key] = value;
+    }
```
sabellius: I think there is inaccuracies in the conditionals here - We what to update an `entity` based on whether an `entityId` was passed and not because of an absence  of an instance of that entity, meaning, you wouldn't want to update a list just because for some reason you couldn't find a card (and the same goes for a board/list).
Doing this:
```js
const card = list.cards?.find(c => c.id === cardId);
```
Is more appropriate to do only if `cardId` was passed as an argument, because if it wasn't your'e iterating through the list of cards needlessly (which is not a big of a deal now but imagine list would've contained many thousands of cards).

Let me know if have any more questions about what I wrote.

sabellius (approved):

PR #7, opened 2025-10-14
feature/CATC_15-add-user-state-to-the-global-state

sabellius:
## Description
This branch adds a reducer and actions for managing user state globally across the app.

## Changes
- Add `.env` files to `gitignore` to avoid pushing local/development environment variables to source control.
- Modify the starter code of the user actions and reducer to be consistent with the conventions delivered in [feature/CATC_16-add-board-state-to-a-global-state](https://github.com/LiorRey/ca-final-project/pull/5), including better error handling, use of action creators.
- Remove some store actions that don't make sense in our app such as increment/decrement.
- Rename user reducer key in the store from `userModule` to `users` until we decide which way we prefer.

### DIFF
--- a/src/store/actions/user-actions.js
+++ b/src/store/actions/user-actions.js
@@ -1,87 +1,105 @@
+import {
+  SET_USERS,
+  SET_USER,
+  SET_WATCHED_USER,
+  DELETE_USER,
+  SET_LOADING,
+  SET_ERROR,
+} from "../reducers/user-reducer";
+
+import { store } from "../store";
 import { userService } from "../../services/user";
 import { socketService } from "../../services/socket-service";
-import { store } from "../store";
-
-import { showErrorMsg } from "../../services/event-bus-service";
-import { LOADING_DONE, LOADING_START } from "../reducers/system-reducer";
-import {
-  REMOVE_USER,
-  SET_USER,
-  SET_USERS,
-  SET_WATCHED_USER,
-} from "../reducers/user-reducer";
 
 export async function loadUsers() {
   try {
-    store.dispatch({ type: LOADING_START });
+    store.dispatch(setLoading(true));
     const users = await userService.getUsers();
-    store.dispatch({ type: SET_USERS, users });
-  } catch (err) {
-    console.log("UserActions: err in loadUsers", err);
+    store.dispatch(setUsers(users));
+  } catch (error) {
+    store.dispatch(setError(`Error loading users: ${error.message}`));
   } finally {
-    store.dispatch({ type: LOADING_DONE });
+    store.dispatch(setLoading(false));
   }
 }
 
-export async function removeUser(userId) {
+export async function deleteUser(userId) {
   try {
     await userService.remove(userId);
-    store.dispatch({ type: REMOVE_USER, userId });
-  } catch (err) {
-    console.log("UserActions: err in removeUser", err);
+    store.dispatch(deleteUserAction(userId));
+  } catch (error) {
+    store.dispatch(setError(`Error deleting user: ${error.message}`));
+    throw error;
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/store/actions/user-actions.js
```diff
@@ -1,87 +1,105 @@
...
+  SET_LOADING,
+  SET_ERROR,
 } from "../reducers/user-reducer";
 
+import { store } from "../store";
+import { userService } from "../../services/user";
+import { socketService } from "../../services/socket-service";
+
+export function setUsers(users) {
+  return { type: SET_USERS, payload: users };
```
snoxlax: Why do we need to export these where are they used other than here?

also I wouldnt create functions for these but its fine if we agree to do it
what @LiorRey thinks?

snoxlax (approved):
Looks good!

src/store/actions/user-actions.js
sabellius: If you're talking about `setUsers` and the other action creators, I've mentioned the reasoning in the pull request - The pros are:
- Decoupling component from action structure.
- Easier modification of the action object structure in the future - changes are made only to the action creator instead of the object literals spread across the app.

Cons:
- There is obviously more boilerplate involved with each additional action.

src/store/reducers/user-reducer.js
```diff
@@ -1,57 +1,36 @@
 import { userService } from "../../services/user";
 
-export const INCREMENT = "INCREMENT";
-export const DECREMENT = "DECREMENT";
-export const CHANGE_COUNT = "CHANGE_COUNT";
+export const SET_USERS = "SET_USERS";
 export const SET_USER = "SET_USER";
 export const SET_WATCHED_USER = "SET_WATCHED_USER";
 export const REMOVE_USER = "REMOVE_USER";
```
LiorRey: I agree with the term **`delete`** rather than `remove` for user, as this means permanent elimination of the user from the storage rather than disassociating him (for example, in MongoDB we use the `db.collection.deleteOne(filter)` method to delete a single document matching the filter from the db collection, so I think this also reinforces the use of the term **`delete`** over `remove`).

So instead of `REMOVE_USER` I think we should use **`DELETE_USER`** and use here in the term **`delete`** anywhere within `user` scope.

src/store/actions/user-actions.js
LiorRey: I read about the concept of **createAction** in Redux, and I actually kind of like it.
Nevertheless, I'd like to hear Tal's opinion on it.

LiorRey:
Other than using `remove` instead of **`delete`**, it looks great to me 👍

LiorRey (approved):

src/store/actions/user-actions.js
snoxlax: I was talking about for example here
we have export of setUser is it needed  cant we just do `function setUser(user)` or am I missing something?
```js
export function setUser(user) {
  return { type: SET_USER, payload: user };
}
```
and here we have another export (this one I understand beacuse we will control it with the frontend)
```js
export async function login(credentials) {
  try {
    const user = await userService.login(credentials);
    store.dispatch(setUser(user));
    socketService.login(user._id);
    return user;
  } catch (error) {
    store.dispatch(setError(`Error logging in: ${error.message}`));
    throw error;
  }
}
```

sabellius: @snoxlax I exported them because I believed that some actions are dispatched directly from the component like we did sometime in the course. I can't find a scenario where it's needed at the moment so I agree that it's not necessary to export those - Remove the exports.

src/store/reducers/user-reducer.js
sabellius: Good point! Done.

PR #8, opened 2025-10-16
fix: Chnaged and fixed styles and code #CATC-22

snoxlax:
#### What I Did

- Fixed Styles and Alignments
- Removed .jsx from imports
- Removed double store implementation
- Removed App.css and main.css (we have it in styles)

**Changed Export Default to Named Exports**

- **Before**: `export default function ComponentName()`
- **After**: `export function ComponentName()`
- **Benefits**: Better tree-shaking, improved code clarity, modern ES6+ practices
- **Files affected**: Some React components

**Changed Icons to Use MUI Icons**

- **Replaced**: Custom icon implementations
- **Added**: Material-UI Icons (`@mui/icons-material`)

**Dependencies changes**

Added

```json
{
  "@mui/icons-material": "^7.3.4",
  "@emotion/react": "^11.14.0",
  "@emotion/styled": "^11.14.1"
}
```

Removed

```json
{
  "lucide-react": "^0.545.0"
}
```

PR #9, opened 2025-10-16
feat: Board interaction menus and responsiveness #CATC-23

snoxlax:
#### What I Did

- Added a **Popover** from MUI to the card for **edit** and **delete** options, so we can later connect it with the interaction and state management.

- The Popover wasn’t as simple as I thought it would be, but it’s **almost fully working now**. We just need to add a few more things to the menu for full card control.

- For the **lists**, I used a simpler and temporary approach with the same Popover right now it only deletes a list, just so we can start testing it with the state management.

- Did some more **CSS adjustments** in general to align things and clean up the layout.

Let me know what you think about the **Popover component** and how it’s implemented if you have other ideas or improvements, I’m open to them!

#### Bonus

Almost done with the **board responsiveness**!
It was super frustrating since I don’t really know CSS that well, but I somehow managed to make the board actually responsive now.
Not entirely sure _why_ it works but it does 😅 I have some idea though, and it’s working pretty well.

### DIFF
--- a/src/components/Card.jsx
+++ b/src/components/Card.jsx
@@ -2,35 +2,82 @@ import {
   ChatRounded,
   NotesRounded,
   RemoveRedEyeOutlined,
+  DriveFileRenameOutline,
 } from "@mui/icons-material";
 
+import CardPopover from "./CardPopover";
+
+import { Box } from "@mui/material";
+import { useState } from "react";
+
 export function Card({ card, onRemoveCard, onUpdateCard }) {
-  function onRemoveCard() {
-    onRemoveCard(card.id);
+  const [anchorEl, setAnchorEl] = useState(null);
+
+  function handleClick(e) {
+    setAnchorEl(e.currentTarget);
   }
 
-  function onUpdateCard() {
-    onUpdateCard(card);
+  function handleClose() {
+    setAnchorEl(null);
   }
+
+  function handleSave() {}
+
+  function handleDelete() {
+    onRemoveCard(card.id);
+    handleClose();
+  }
+
+  const open = Boolean(anchorEl);
+  const id = open ? `card-popover` : undefined;
 
   return (
     <section className="card-container">
-      <div className="card-labels">
-        {card.labels.map(label => (
-          <div
-            key={`${card.id}-${label}`}
-            className={`card-label ${label.color}`}
-          >
-            {label.name}
+      <Box
+        className={`${open ? "floating-card-content" : "card-content"}`}
+        sx={open ? { zIndex: theme => theme.zIndex.modal + 1 } : {}}
+      >
+        <div className="card-labels">
+          {card.labels.map(label => (
+            <div
+              key={`${card.id}-${label.name}-${label.color}`}
+              className={`card-label ${label.color}`}
+            >
+              {label.name}
+            </div>
+          ))}
+        </div>
+        <h3 className="card-title">{card.title}</h3>
+        {open ? (
+          <div className="card-footer">
+            <div className="empty-card-footer" />
           </div>
-        ))}
-      </div>
-      <h3 className="card-title">{card.title}</h3>
-      <div className="card-footer">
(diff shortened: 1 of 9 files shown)


snoxlax:
This pr also contains the changes from previous pr [this one](https://github.com/LiorRey/ca-final-project/pull/8) and it will be merged after that one is merged

I will try and think how can I make it so only the relevant changes to this pr are only visible if you know how to do it tell me

Edit: I changed the base branch to this [this one](https://github.com/LiorRey/ca-final-project/pull/8) for now

LiorRey (approved):
Looks very promising, would really like to see the magic!🧙‍♂️

src/components/Card.jsx
```diff
@@ -2,35 +2,82 @@ import {
...
 
+import CardPopover from "./CardPopover";
+
+import { Box } from "@mui/material";
+import { useState } from "react";
+
 export function Card({ card, onRemoveCard, onUpdateCard }) {
-  function onRemoveCard() {
-    onRemoveCard(card.id);
+  const [anchorEl, setAnchorEl] = useState(null);
```
sabellius, line 14: Isn't that a good place to use a ref?

src/components/Card.jsx
```diff
@@ -2,35 +2,82 @@ import {
...
-      <div className="card-labels">
-        {card.labels.map(label => (
-          <div
-            key={`${card.id}-${label}`}
-            className={`card-label ${label.color}`}
-          >
-            {label.name}
+      <Box
+        className={`${open ? "floating-card-content" : "card-content"}`}
+        sx={open ? { zIndex: theme => theme.zIndex.modal + 1 } : {}}
```
sabellius, line 38: I've read that the `sx` property is used to define custom css for a component. But using arbitary `zIndex` value can lead to issues later. Do the docs tell what default `zIndex` values are used by default? it's usually better to work with those values if those exist. for Reference here is an example from Bootstrap docs: https://getbootstrap.com/docs/5.3/layout/z-index/

src/components/Card.jsx
```diff
@@ -2,35 +2,82 @@ import {
...
+          </div>
+        )}
+        <button
+          onClick={handleClick}
+          aria-describedby={id}
+          aria-label="Edit card"
+          className="icon-button card-edit-button"
+        >
+          <DriveFileRenameOutline />
+        </button>
```
sabellius, line 69: Since were using material ui, I assume they have a button component - why not use it? This also applies to other ui components as well.

sabellius (changes requested):

src/components/Card.jsx
snoxlax: I tried to do that actually with their button but it was to customized I think it would take more code to remove their CSS and customize it to what we need then to just use our button

src/components/Card.jsx
snoxlax: I think its the same here
https://mui.com/material-ui/customization/z-index/

So this Box should be above the Backdrop so basiclly it modifies this zIndex to be above the backdrop how would you suggest doing this differently? the numbers should be diffrent? 
also I dont think its modified permanently is it?

src/components/Card.jsx
snoxlax: I'm not sure because it changes if its open that it has a ref but when its closed its back to null 
ill try it with ref and see if its better

src/components/Card.jsx
sabellius: No, I just wanted to check that I understand whats going on. Looks good to me!

src/components/Card.jsx
sabellius: Alrighty then

src/components/Card.jsx
snoxlax: You are saying together with useState or instead of useState?

sabellius (approved):

PR #10, opened 2025-10-17
bugfix/CATC_29-fix-error-crashing-the-app

sabellius:
## Description
Fix error that is crashing the app.

## Changes
- Fix occurrences where `removeUser` is used and changed it to `deleteUser`.
- Fix occurrences where `userModule` is used and changed it to `user`.
- Change user state key in the user reducer from `user` to `currentUser`.

## Notes
- The error was caused to a name change in the starter code.
- I've scanned the entire code base for the problematic names.

### DIFF
--- a/src/pages/AdminIndex.jsx
+++ b/src/pages/AdminIndex.jsx
@@ -1,6 +1,6 @@
 import { useEffect } from "react";
 import { useSelector } from "react-redux";
-import { loadUsers, removeUser } from "../store/actions/user-actions";
+import { loadUsers, deleteUser } from "../store/actions/user-actions";
 import { useNavigate } from "react-router";
 
 export function AdminIndex() {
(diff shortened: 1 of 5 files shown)


src/pages/AdminIndex.jsx
```diff
@@ -1,14 +1,14 @@
...
 
 export function AdminIndex() {
   const navigate = useNavigate();
 
-  const user = useSelector(storeState => storeState.userModule.user);
-  const users = useSelector(storeState => storeState.userModule.users);
-  const isLoading = useSelector(storeState => storeState.userModule.isLoading);
+  const user = useSelector(storeState => storeState.user.user);
+  const users = useSelector(storeState => storeState.user.users);
+  const isLoading = useSelector(storeState => storeState.user.isLoading);
```
snoxlax, line 12: Only this was a sticking point with user and users what we talked about

sabellius: See the updated code and PR description please.

snoxlax (approved):
Yes lets go looks good!

PR #12, opened 2025-10-19
fix: UI fixes and alignments #CATC-30

snoxlax:
#### Changes Made
- Fixed various UI inconsistencies across components  
- Added additional color options for tags for better visual distinction  
- Improved spacing and alignments for a cleaner layout  
- Moved label-related classes into a dedicated CSS file for better organization and maintainability 
- Temporarily disabled socket setup to prevent connection errors

### DIFF
--- a/src/components/Card.jsx
+++ b/src/components/Card.jsx
@@ -37,17 +37,21 @@ export function Card({ card, onRemoveCard, onUpdateCard }) {
         className={`${open ? "floating-card-content" : "card-content"}`}
         sx={open ? { zIndex: theme => theme.zIndex.modal + 1 } : {}}
       >
-        <div className="card-labels">
-          {card.labels.map(label => (
-            <div
-              key={`${card.id}-${label.name}-${label.color}`}
-              className={`card-label ${label.color}`}
-            >
-              {label.name}
-            </div>
-          ))}
-        </div>
-        <h3 className="card-title">{card.title}</h3>
+        {card.labels.length > 0 && (
+          <div className="card-labels">
+            {card.labels.map(label => (
+              <div
+                key={`${card.id}-${label.name}`}
+                className={`card-label ${label.color}`}
+              >
+                {label.name}
+              </div>
+            ))}
+          </div>
+        )}
+        <h3 className={`card-title ${card.labels.length === 0 ? "mr-2" : ""}`}>
+          {card.title}
+        </h3>
         {open ? (
           <div className="card-footer">
             <div className="empty-card-footer" />
--- a/src/services/socket-service.js
+++ b/src/services/socket-service.js
@@ -16,9 +16,13 @@ const SOCKET_EMIT_LOGOUT = "unset-user-socket";
 
 const baseUrl = process.env.NODE_ENV === "production" ? "" : "//localhost:3030";
 
+// Temporarily disabled socket setup to prevent connection errors
+/*
 export const socketService =
-  VITE_LOCAL === "true" ? createDummySocketService() : createSocketService();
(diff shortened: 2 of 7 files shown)


src/components/Card.jsx
```diff
@@ -37,17 +37,21 @@ export function Card({ card, onRemoveCard, onUpdateCard }) {
...
-            <div
-              key={`${card.id}-${label.name}-${label.color}`}
-              className={`card-label ${label.color}`}
-            >
-              {label.name}
-            </div>
-          ))}
-        </div>
-        <h3 className="card-title">{card.title}</h3>
+        {card.labels.length > 0 && (
```
sabellius, line 40: It doesn't really matter, but in Javascript `0` is a falsy value, so it's not necessary to check if it's greater than `0`.

src/components/Card.jsx
```diff
@@ -37,17 +37,21 @@ export function Card({ card, onRemoveCard, onUpdateCard }) {
...
+          <div className="card-labels">
+            {card.labels.map(label => (
+              <div
+                key={`${card.id}-${label.name}`}
```
sabellius, line 44: I think we should add an `id` property to each label. I'm assuming you can create multiple labels with the same name, which will cause key-related issues (because of identical keys in map()). Also should a card contain the labels? Aren't labels a board-wide entity?

src/services/socket-service.js
```diff
@@ -16,9 +16,13 @@ const SOCKET_EMIT_LOGOUT = "unset-user-socket";
...
 const baseUrl = process.env.NODE_ENV === "production" ? "" : "//localhost:3030";
 
+// Temporarily disabled socket setup to prevent connection errors
+/*
 export const socketService =
-  VITE_LOCAL === "true" ? createDummySocketService() : createSocketService();
+   VITE_LOCAL === "true" ? createDummySocketService() : createSocketService();
+ */
 
+export const socketService = createDummySocketService();
```
sabellius, line 25: Why those changes are needed? assuming the starter code works, and there is a behavior for development environments, what errors did you get?

sabellius (changes requested):

src/components/Card.jsx
snoxlax: Yes ids for label I was thinking about this too they will also appear in other places so they need ids too
We need to think where to put them maybe inside the board object?

src/services/socket-service.js
snoxlax: Thats beacuse there is an annoying error in the dev tools console I just disabled it for now so it wont flood the console
cuase the socketService is trying to connect and its retrying over and over again
check it out in devtools console

src/components/Card.jsx
snoxlax: Yes but 0 && true is false and then it returns 0 and it renders the 0 in the UI

snoxlax: I can do if you want

` {!!card.labels?.length && (`

sabellius: Oh never mind, I forgot about this React pitfall, my bad.

src/components/Card.jsx
sabellius: In a relational database, there would be a junction table that connects cards with lables with foreign keys (`cardId`, `LabelId`). But since we're going to use a non-relational db (MongoDB), perhaps we should put label on the top level of the board and reference them by IDs in the card.
I've created a [ticket](https://trello.com/c/68Z3xJ2Z/34-refactor-normalize-label-storage-to-eliminate-duplication) for this because it's not the scope of this PR. However you need to use some unique value here.

sabellius (approved):

PR #13, opened 2025-10-19
feat: Added edit card from the card itself #CATC-35

snoxlax:
#### What I Did

- Added the ability to **edit the card title directly from the card** by clicking the **edit button** on the top-right corner of the card.
- While editing the title:
  - Pressing **Save** will save the updated title **locally** (for now).
  - Clicking **anywhere outside** the editing area will **exit the edit mode without saving**, discarding changes.

## Notes

- The editing UI is inline and replaces the static title temporarily.
- The feature currently stores changes **only locally**

### DIFF
--- a/src/assets/styles/components/Card.css
+++ b/src/assets/styles/components/Card.css
@@ -21,6 +21,9 @@
 }
 
 .floating-card-content {
+  display: inline-block;
+  position: relative;
+  overflow: visible;
   display: flex;
   flex-direction: column;
   border-radius: 6px;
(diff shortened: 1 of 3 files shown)


LiorRey (approved):
Looks good, and after we've dicussed, settled things and ensured there are no conflicts anymore around the whole add/edit card flow, I approve 🦾

sabellius (approved):
Looks good! Do we have a ticket for connecting this to the global state?

snoxlax:
This is just the visual part only the UI the other part of actually updating the global state is here #15

PR #14, opened 2025-10-19
feature/CATC_27-allow-users-to-add-a-list-to-a-board

sabellius:
## Description
This PR add functionality to allow users to create a new list directly from the board.

## Changes
- `UPDATE_BOARD` action also updates the board property in the store so that components that subscribe to updates of the board could re-render.
- `BoardDetails` not gets board state directly from the store instead of props. The page structure is wrong and needs to be fixed in a separate issue (`BoardIndex` renders `BoardDetails`).
- Add a simple AddForm component that uses local state to keep track of the list name, and on submit, updates the board lists array.

## Code review changes:
- Load boards using store actions in the `BoardIndex` compnent, and immediately navigate to the first boards details page.
-  Use `usParams` and load the board in `BoardDetails` from the store using the id from params.
- In the board-reducer, avoid updating the boards array in the `UPDATE_BOARD` actions, and instead, update the board object.

## Demo
[Screencast from 10-19-2025 06:56:59 PM.webm](https://github.com/user-attachments/assets/e1a88145-2ec9-4b4c-bdc0-c7920a9e48b8)

### DIFF
--- a/src/components/AddList.jsx
+++ b/src/components/AddList.jsx
@@ -1,0 +1,54 @@
+import { useState } from "react";
+
+import IconButton from "@mui/material/IconButton";
+import Button from "@mui/material/Button";
+import TextField from "@mui/material/TextField";
+import AddIcon from "@mui/icons-material/Add";
+import CloseIcon from "@mui/icons-material/Close";
+
+import { boardService } from "../services/board";
+
+export function AddList({ onSubmit }) {
+  const [showAddList, setShowAddList] = useState(true);
+  const [listName, setListName] = useState("");
+
+  function onSubmitAddList() {
+    if (!listName) return;
+    const newList = boardService.getEmptyList();
+    newList.name = listName;
+    onSubmit(newList);
+    setListName("");
+    setShowAddList(true);
+  }
+
+  return (
+    <>
+      {showAddList && (
+        <Button
+          startIcon={<AddIcon />}
+          className="add-list-button"
+          onClick={() => setShowAddList(false)}
+        >
+          Add another list
+        </Button>
+      )}
+      {!showAddList && (
+        <div className="add-list-container">
+          <div>
+            <TextField
+              id="outlined-basic"
+              variant="outlined"
+              value={listName}
+              placeholder="Enter list name"
+              onChange={e => setListName(e.target.value)}
+            />
+          </div>
+          <Button onClick={onSubmitAddList}>Add List</Button>
+          <IconButton aria-label="close" onClick={() => setShowAddList(true)}>
+            <CloseIcon />
+          </IconButton>
+        </div>
+      )}
+    </>
+  );
+}
--- a/src/pages/BoardDetails.jsx
+++ b/src/pages/BoardDetails.jsx
@@ -1,20 +1,16 @@
-import { AddRounded } from "@mui/icons-material";
-import {
-  MoreHoriz,
-  Sort,
-  StarBorderRounded,
-  LockOutlineRounded,
-} from "@mui/icons-material";
+import MoreHoriz from "@mui/icons-material/MoreHoriz";
+import Sort from "@mui/icons-material/Sort";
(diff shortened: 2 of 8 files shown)


src/pages/BoardDetails.jsx
```diff
@@ -1,20 +1,29 @@
...
-import { AddRounded } from "@mui/icons-material";
+import { AddList } from "../components/AddList";
 
-import {
-  MoreHoriz,
-  Sort,
-  StarBorderRounded,
-  LockOutlineRounded,
-} from "@mui/icons-material";
+export function BoardDetails() {
```
snoxlax, line 15: So I think this one should be with useParams() cuase if you remember we are doing 
`<Route path="board/:boardId" element={<BoardDetails />} />`
but I dont mind how it is right now if we just need it to work

src/store/reducers/board-reducer.js
```diff
@@ -28,7 +28,16 @@ export function boardReducer(state = initialState, action) {
       const updatedBoards = state.boards.map(board =>
         board._id === action.payload._id ? action.payload : board
       );
-      return { ...state, boards: updatedBoards };
+      const updatedCurrentBoard =
```
snoxlax: Shouldnt we put this in actions? it looks a bit complex

src/pages/BoardIndex.jsx
```diff
@@ -1,24 +1,21 @@
...
 
 export function BoardIndex() {
-  const [board, setBoard] = useState(null);
-
   useEffect(() => {
     fetchBoard();
   }, []);
 
   async function fetchBoard() {
     const boards = await boardService.query();
```
snoxlax, line 23: Also here we should not use service I think we should get the boards from redux no?

snoxlax (changes requested):
Other than what I wrote looks good!

src/pages/BoardIndex.jsx
LiorRey: We should use loadBoards, or initially loadBoard(boardId) here, rather than using boardService in the Index.

src/store/reducers/board-reducer.js
sabellius: I think I should just update the board property here and not touch the board array.

src/pages/BoardIndex.jsx
sabellius: I now get the board using the `params.boardId` and load it to the store and use it from there.

src/components/AddList.jsx
```diff
@@ -0,0 +1,54 @@
...
+    newList.name = listName;
+    onSubmit(newList);
+    setListName("");
+    setShowAddList(true);
+  }
+
+  return (
+    <>
+      {showAddList && (
+        <Button
```
snoxlax, line 27: The styling looks off because its using an MUI Button.

Like I mentioned before, we might not need to rely on some of these components we could build our own custom ones ourselves.

That would make the design less dependent on the library among other things.

src/services/board/board-service-local.js
```diff
@@ -78,6 +79,14 @@ async function updateBoardWithActivity(
   }
 }
 
+export function getEmptyList() {
```
LiorRey, line 82: Shouldn't it be in the `services/board/index.js`?

src/pages/BoardDetails.jsx
sabellius: Done

src/components/AddList.jsx
sabellius: I mostly agree, there are some benefits using MUI components as they handle various things behind the scenes, but I also not too familiar with how easy is (or isn't) is to customize them. Regardless as we discussed in Discord, I'm keeping it ugly as it is now to focus more on functionality.

snoxlax (approved):
Looks great!

LiorRey (approved):
After we've discussed and settled things on Discord - I approve 🦾

PR #15, opened 2025-10-20
feat: Added save changes of edited card #CATC-37

snoxlax:
### In-Card Editing Save

**Description:**  
Added the functionality to save edits made directly within a card.

**Notes:**  
- These changes depend on the previous visual editing branch #13
- Also depends on @sabellius's branch for adding a new list #14
- This branch should be merged **after** the above branches are merged.

### DIFF
--- a/src/pages/BoardDetails.jsx
+++ b/src/pages/BoardDetails.jsx
@@ -12,9 +12,6 @@ import {
 export function BoardDetails({ board }) {
   async function onRemoveList(listId) {}
 
-<<<<<<< Updated upstream
-  async function onAddList() {}
-=======
   async function onUpdateList(list) {
     updateBoard(board, {
       listId: list.id,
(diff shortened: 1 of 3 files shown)


src/pages/BoardDetails.jsx
```diff
@@ -12,7 +12,13 @@ import {
 export function BoardDetails({ board }) {
   async function onRemoveList(listId) {}
 
-  async function onAddList() {}
+  async function onUpdateList(list) {
```
LiorRey, line 15: Shouldn't it be what we discussed on Discord?
I mean this:
```
  async function onUpdateList(list, { cardId = null, key, value }) {
    try {
      const options = { listId: list.id, cardId, key, value };
      updateBoard(board, options);
      showSuccessMsg(`The list ${list.name} updated successfully!`);
    } catch (error) {
      console.error("List update failed:", error);
      showErrorMsg(`Unable to update the list: ${list.name}`);
    }
  }
```

And there should be an empty line between `onUpdateList` and `onSubmitAddList` (saw these functions one after another without any space on `main` branch, just wanted to make sure they will be separated).

LiorRey (approved):
After we've discussed and settled it on Discord, I approve 🦾

PR #20, opened 2025-10-21
feat: Add horizontal scroll to end when engaging the AddList component

LiorRey:
Add horizontal scroll to end when engaging the AddList component


src/pages/BoardDetails.jsx
```diff
@@ -37,6 +38,24 @@ export function BoardDetails() {
...
+        const el = boardCanvasRef.current;
+        if (el) {
+          const scrollTarget = el.scrollWidth - el.clientWidth + 100;
+          el.scrollTo({
+            left: scrollTarget,
+            behavior: "smooth",
+          });
+        }
+      },
+      isAfterListAdd ? 500 : 0
```
sabellius: I think it's a good idea to put this is a variable with a meaningful name that tells that this value equals "delay"

src/components/AddList.jsx
```diff
@@ -44,11 +44,18 @@ export function AddList({ onSubmit }) {
...
+            onClick={() => {
+              setShowAddList(true);
+              setListName("");
+            }}
+          >
             <CloseIcon />
           </IconButton>
         </div>
       )}
+      {!showAddList && scrollHorizontallyBoardCanvasToEnd()}
```
sabellius: What is this doing? Regardless, seems like it's something that should be done in a `useEffect` maybe?

src/components/AddList.jsx
```diff
@@ -44,11 +44,18 @@ export function AddList({ onSubmit }) {
             />
           </div>
           <Button onClick={onSubmitAddList}>Add List</Button>
-          <IconButton aria-label="close" onClick={() => setShowAddList(true)}>
+          <IconButton
+            aria-label="close"
+            onClick={() => {
```
sabellius: I think once it's not a one-liner anymore, it's better to extract it to a regular function. What do you think?

sabellius (changes requested):
Please read my comments and questions.

src/components/AddList.jsx
snoxlax: `scrollHorizontallyBoardCanvasToEnd()`
can we not use AI to name our function please?😅

snoxlax (changes requested):
Looks good to me!
The only thing is I don’t think this should be inside BoardDetails.

Maybe it could be moved to a separate custom hook or logic file?
I just don’t think the implementation should live inside BoardDetails.

if its not possible or we dont know how to do it right now tell me and i will approve.

src/pages/BoardDetails.jsx
LiorRey: Done 🦾
```
    const DELAY_AFTER_LIST_ADD = 500;
    const NO_DELAY = 0;
    const delay = isAfterListAdd ? DELAY_AFTER_LIST_ADD_MS : NO_DELAY;

......
    }, delay);
```

src/components/AddList.jsx
LiorRey: @sabellius 
It was done quickly just before the presentation, it supposed to slide the scroll the board to its end (rightmost side).
Fixed it, now done in a `useEffect`. 👍

@snoxlax 
Actually it was renamed by me, preferred at first a long and self-explaining name for this function.
Fixed it, now it's called `scrollBoardToEnd()` .👍

src/components/AddList.jsx
LiorRey: Also this was done rashly for the presentation.
Fixed it back to one-liner. 👍

src/components/AddList.jsx
```diff
@@ -8,43 +8,65 @@ import CloseIcon from "@mui/icons-material/Close";
...
+
+    const newList = boardService.getEmptyList();
+    newList.name = newListName;
+    onAddList(newList);
+
+    setNewListName("");
+    textFieldRef.current?.focus();
+  }
+
+  const handleCloseAddListActions = () => {
```
snoxlax: Long name maybe handleCloseAddList?

src/components/List.jsx
```diff
@@ -7,32 +7,42 @@ import Popover from "@mui/material/Popover";
...
-    setOpenModal(false);
-    setSelectedCard(null);
-  }
+  useEffect(() => {
+    if (isAddingCard) {
+      newCardInputRef.current?.focus();
+    }
+  }, [isAddingCard]);
+
+  useEffectUpdate(() => {
```
snoxlax: Why do we need this hook?

src/components/List.jsx
```diff
@@ -62,20 +72,21 @@ export function List({ list, boardLabels, onRemoveList, onUpdateList }) {
...
+    onSetAddingCard(null);
   };
 
-  function handleAddCardTitleChange({ target }) {
-    setNewCardTitle(target.value);
-  }
-
   const handleAddCard = () => {
+    if (!newCardTitle) {
+      handleHidingAddCardActions();
```
snoxlax: handleHidingAddCardActions is too long why do we need 'Actions' also we cant use Hiding we need clear action handleOpen/handleClose handleShow/handleHide

src/components/List.jsx
```diff
@@ -156,12 +168,12 @@ export function List({ list, boardLabels, onRemoveList, onUpdateList }) {
...
               <input
+                ref={newCardInputRef}
                 type="text"
-                placeholder="Enter a title or paste a link"
                 className="card-title-input"
-                autoFocus
                 value={newCardTitle}
-                onChange={handleAddCardTitleChange}
+                onChange={e => setNewCardTitle(e.target.value)}
+                placeholder="Enter a title or paste a link"
```
snoxlax, line 166: Why the change here? this code is the section for the floating card from the Card component why is it here? 😅
how is this part of the code of scroll to end

src/pages/BoardDetails.jsx
```diff
@@ -14,16 +14,34 @@ import { FilterMenu } from "../components/FilterMenu";
...
+  const didLoadOnce = useRef(false);
+  const boardCanvasRef = useRef(null);
+  const scrollBoardToEnd = useScrollTo(boardCanvasRef);
 
   useEffect(() => {
+    didLoadOnce.current = false;
     loadBoard(params.boardId);
   }, [params.boardId]);
 
+  useEffectUpdate(() => {
```
snoxlax: The same about this hook why do we need it?

src/pages/BoardDetails.jsx
```diff
@@ -14,16 +14,34 @@ import { FilterMenu } from "../components/FilterMenu";
 import { showErrorMsg, showSuccessMsg } from "../services/event-bus-service";
 import { getFilteredBoard } from "../services/filter-service";
 import { useCardFilters } from "../hooks/useCardFilters";
+import { SCROLL_DIRECTION, useScrollTo } from "../hooks/useScrollTo";
+import { useEffectUpdate } from "../hooks/useEffectUpdate";
 
 export function BoardDetails() {
+  const [activeAddCardListId, setActiveAddCardListId] = useState(null);
```
snoxlax, line 24: ?

src/components/AddList.jsx
LiorRey: By "actions" I meant the part that opens when clicking on "Add another list" button :
image

But I changed to handleCloseAddList 👍

src/components/List.jsx
LiorRey: This is a custom hook that already was in `hooks` folder, and it works like useEffect, but skips the first render.
We need it to prevent the lists from scrolling all the way down to the end after reloading the page.
If I use regular `useEffect`, it does not handle the first render, which is what we need.

src/components/List.jsx
LiorRey: Fixed

sabellius (commented):
test

LiorRey (commented):
Handle pending comment

src/pages/BoardDetails.jsx
LiorRey: Fixed it 🦾

src/components/List.jsx
LiorRey: In our Discord discussion, after seeing that `handleAddCardTitleChange()` is a function with one line only calling `setNewCardTitle()` - we decided that `onChange={e => setNewCardTitle(e.target.value)}` is better 🦾

About the section of the floating card from Card component -
Now fixed 🦾

src/components/List.jsx
LiorRey: Fixed it 🦾

src/components/AddList.jsx
LiorRey: Fixed it 🦾

src/pages/BoardDetails.jsx
LiorRey: I tried different ways, but only the current way of using this state in `BoardDetails.jsx`:
`const [activeAddCardListId, setActiveAddCardListId] = useState(null);`
and then :
```
 {filteredBoard.lists.map(list => (
            <li key={list.id}>
              <List
                key={list.id}
                list={list}
                boardLabels={board.labels}
                onRemoveList={onRemoveList}
                onUpdateList={onUpdateList}
                isAddingCard={activeAddCardListId** === list.id}
                setActiveAddCardListId={setActiveAddCardListId}
              />
            </li>
          ))}
```
          
without any additional state or extra checks, can ensure that **only one list at a time can have its "add card" form open**.

sabellius (approved):
Looks good!

LiorRey:
All agreed to merge, even if it is not shown here

LiorRey (commented):
fixed and done

PR #21, opened 2025-10-22
feature/CATC_34-normalize-label-storage-storage-to-eliminate-duplication

sabellius:
## 📝 Description

Refactored board architecture to use board-level labels with normalized card references, eliminating code duplication.

## 🔧 Changes

- __Data Structure__: Moved labels to board level, cards now reference by ID.
- __Components__: Updated BoardDetails → List → Card to pass and resolve labels.

## 📋 Notes

- Eliminates duplicate label arrays across functions.
- Cards use label IDs instead of full objects for better performance.
- Single source of truth for label definitions.
- The visuals look exactly the same.

### DIFF
--- a/src/components/List.jsx
+++ b/src/components/List.jsx
@@ -101,16 +101,25 @@ export function List({ list, boardLabels, onRemoveList, onUpdateList }) {
       </div>
       <div className="list-content-container" ref={listContentRef}>
         <ul className="cards-list">
-          {cards.map(card => (
-            <li key={card.id}>
-              <Card
-                key={card.id}
-                card={card}
-                onRemoveCard={onRemoveCard}
-                onUpdateCard={onUpdateCard}
-              />
-            </li>
-          ))}
+          {cards.map(card => {
+            const cardLabels =
+              boardLabels && card.labels && card.labels.length > 0
+                ? card.labels
+                    .map(labelId => boardLabels.find(l => l.id === labelId))
+                    .filter(Boolean)
+                : [];
+            return (
+              <li key={card.id}>
+                <Card
+                  key={card.id}
+                  card={card}
+                  labels={cardLabels}
+                  onRemoveCard={onRemoveCard}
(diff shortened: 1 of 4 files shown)


src/components/List.jsx
```diff
@@ -101,16 +101,25 @@ export function List({ list, onRemoveList, onUpdateList }) {
...
+                ? card.labels
+                    .map(labelId => boardLabels.find(l => l.id === labelId))
+                    .filter(Boolean)
+                : [];
+            return (
+              <li key={card.id}>
+                <Card
+                  key={card.id}
+                  card={card}
+                  labels={cardLabels}
```
snoxlax, line 116: This works but what about if we have users assigned to the card for example than we will have a users array sent to the card as well right?

then we have the Modal of the CardDetails we would need to send it there too the Labels array and Users array

so maybe we could have it integrated inside card object somehow after we find what labels the card has and what users are assigned to the card we sort of populate the card object instead of label ids we put in the real labels etc.

hope its clear. maybe there is a better solution.

snoxlax (approved):
Looks good!
Other than what I wrote maybe we could make it more future proof

src/components/List.jsx
sabellius: This is part of the question that I've asked Tal, but not because of sending props that the card "already has", but because of the constant calculations and re-rendering this will causes since everything is "part of the board".
The only solutions for not passing referenced information to cars such labels, members, etc.. I can think of are:
- Resolve this information on the server and send the response with complete information and without referencing. This method brings back redundancy (multiple copies of the same objects will be returned in the response) and maybe other bad side-effects.
- Make the card component a "smart" component that can access the board from the store and get all the information it needs itself. This is an easier solution but it's not very in the "React way".

If you have other solutions other than modifying the card objects themselves which is usually a big red flag in React, I'd love to hear them.

LiorRey (approved):
After we've covered the logic and code of the labels storage, I approve 🦾

PR #22, opened 2025-10-22
feature/CATC_42-build-a-filter-menu-ui-component

sabellius:
## **🎯 Description**
Built a comprehensive card filtering system with UI component, state management, and performance optimizations.

## **🔧 Changes**
- **Added FilterMenu UI component** with Material-UI popover and search input
- **Added useMemo optimization** to prevent unnecessary re-filtering on renders
- **Implemented debounced filtering** to avoid filtering on every keystroke

## **📝 Notes**
- Complete filtering system from state management to UI implementation
- Performance optimized with memoization and debouncing techniques  
- Uses existing util-service debounce function for consistency

### DIFF
--- a/src/components/FilterMenu.jsx
+++ b/src/components/FilterMenu.jsx
@@ -1,4 +1,4 @@
-import { useState, useRef } from "react";
+import { useState, useRef, useEffect } from "react";
 import Popover from "@mui/material/Popover";
 import IconButton from "@mui/material/IconButton";
 import Box from "@mui/material/Box";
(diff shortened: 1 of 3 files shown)


snoxlax (approved):
Looks good to me.

About the use of MUI components like Typography I get that it’s good for accessibility (ARIA and all), but the inline sx prop makes it harder to customize styling.

Is it necessary to use MUI here, or could we make some of these simpler components ourselves instead?

sabellius:
> Looks good to me.
> 
> About the use of MUI components like Typography I get that it’s good for accessibility (ARIA and all), but the inline sx prop makes it harder to customize styling.
> 
> Is it necessary to use MUI here, or could we make some of these simpler components ourselves instead?

It was a suggestion of ai to use the Typography component, and the reason I chose to keep it, besides the accessibility features, it standartizes how text is displayed across the app (font family, weight, size, etc..) without having to write many inconsistent css classes to style regular html elements. This is in my opinion a better solution than having many un-organized css classes floating around in the project that are used inconsistently. 

Regardless of the Typography component, MUI components can be styled using regular css classes as well as other ways like the `sx` property you mentioned. How to do this is mentioned [here](https://mui.com/material-ui/customization/how-to-customize/#overriding-styles-with-class-names) and [here](https://mui.com/material-ui/integrations/interoperability/#plain-css).

Perhaps we need to re-discuss the use of MUI in the project, and if we decide to use it, which components we would like to use and which we prefer to write our own from scratch.

My personal opinion is that the best of both worlds, for example buttons is to wrap the mui component with our own, and give it different styles using the ways mentioned in the docs. This will allow us to keep using components instead of looking for the correct css classes across our stylesheets.

LiorRey:
After watching about the use of it in the last lecture again ("Lesson10 - ReactAwesome - Part1", time 52:00 onward) and especially after using it in my PRs, I tend to like and get along with MUI.
I think it will be great if we'll use it in general, and of course only the controls that we need like Box, AddIcon, TextField, IconButton, the contained button and so on.

PR #23, opened 2025-10-24
feat: Added the functionality of the CardDetails page #CATC-50

snoxlax:
## 🎯 Description

This PR implements the **basic functionality of the CardDetails modal page**.
It allows opening the modal, editing a card’s details, and saving the changes.

This also now includes deleting the card from the modal

Currently, the CardDetails component **opens as its own page** but still **retains the modal-like appearance** for consistency.

## 🔧 Changes

* Added a route for the CardDetails page.
* Updated navigation logic to handle CardDetails as a standalone page.
* Ensured modal-style visuals remain consistent even as a full page.
* Added remove card from the modal
* Refactored the whole card handling including Add, Edit and Delete
* In the board-service-local I also refactored the temp code that handles the logic of updating the board

## 📝 Notes
Still a work in progress you can already check the overall structure, behavior, and logic flow for early feedback.

I’ve implemented the new changes, and we’re now working with the refactored structure for updating cards through Redux. This includes adding, editing, and removing cards.

Next, we’ll need to gradually refactor the other instances of these actions to align with the new structure.

## 🖥️ Demo
coming soon


src/components/CardModal.jsx
```diff
@@ -21,7 +22,7 @@ export function CardModal({
   if (!card) return null;
 
   return (
-    <Modal open={isOpen} onClose={onClose} disableAutoFocus>
+    <Modal open={isOpen} onClose={onClose}>
       <Box className={`card-modal-box ${openSection ? "open" : "closed"}`}>
         <div className="card-modal-header">
           {listTitle}{" "}
```
LiorRey: Not big issue but I prefer :
```
{`${listTitle} `}
```
Instead of :
```
{listTitle}{" "}
```

LiorRey (approved):

src/services/board/board-service-local.js
```diff
@@ -145,8 +146,15 @@ function _applyBoardUpdate(
...
-      card[key] = value;
+      if (!key && value) {
+        const index = list.cards.findIndex(c => c.id === cardId);
+        prevValue = card;
+        let newcard = { ...card, ...value };
+        list.cards[index] = newcard;
+      } else {
+        prevValue = card[key];
+        card[key] = value;
+      }
```
sabellius: I think we should send an object instead of a key/value in order to be able to update multiple properties or entire objects instead of a single property. Also, I think we should turn off creating activities as the current implementation asks for information that isn't used by the front-end, as well as activities should be used as an audit-trail system or a recovery system. Activities should just e activities - describe which user, did what, and when, and in which context. @LiorRey @snoxlax

sabellius: * Turn them off until we specify the correct behavior. Regardless I think activities should not be coupled with a single function as they may be relevant to actions that aren't related to boards.

src/App.jsx
```diff
@@ -12,24 +12,23 @@ import { AdminIndex } from "./pages/AdminIndex";
 import { LoginSignup, Login, Signup } from "./pages/LoginSignup";
 
 export function App() {
+  const location = useLocation();
+  const backgroundLocation = location.state?.backgroundLocation;
```
sabellius, line 16: Why is this needed?

sabellius (approved):

sabellius (approved):
Just delete the commented out code. Other than that, looks very promising. Good job!

sabellius (approved):

LiorRey (approved):
After our review of the code in discord discussion, I approve 🦾

PR #24, opened 2025-10-24
feature/CATC_38-implement-members-filtering

sabellius:
## 🎯 Description
This PR introduces members filtering to the board and refactors how board filtering is handled for improved clarity and maintainability.

## 🔧 Changes
- Added the ability to filter cards by assigned members, making it easier to focus on relevant tasks.
- Refactored the board filtering logic: filtering is now performed when loading the board, rather than inside the `BoardDetails` component.

## Notes
- I've added a functionality for the "Assigned to me" filter which currently uses a placeholder for the current users' id and doesn't really work right now, but will require a small change to work once we will introduce a current user state with authentication in the future.

## 🖥️ Demo

<details>
  <summary>Click To Expand</summary>

  <p align="center">
    
[Screencast from 10-24-2025 07:57:14 PM.webm](https://github.com/user-attachments/assets/2cb758af-ade3-4a3e-85e7-461463737897)

  </p>

</details>

### DIFF
--- a/src/assets/styles/components/FilterMenu.css
+++ b/src/assets/styles/components/FilterMenu.css
@@ -1,0 +1,14 @@
+.filter-member-autocomplete .MuiAutocomplete-popupIndicator {
+  transform: none;
+  transition: none;
+}
+
+.filter-member-fullname {
+  font-size: 0.85em;
+}
+
+.filter-member-username {
+  color: #888;
+  font-size: 0.6em;
+  margin-left: 4px;
+}
--- a/src/components/FilterMenu.jsx
+++ b/src/components/FilterMenu.jsx
@@ -22,8 +22,13 @@ export function FilterMenu() {
   const [isOpen, setIsOpen] = useState(false);
   const [localTitle, setLocalTitle] = useState("");
   const anchorRef = useRef(null);
-  const { filters, updateFilterDebounced, updateFilter, clearAllFilters } =
-    useCardFilters();
+  const {
+    filters,
+    updateFilterDebounced,
+    updateFilter,
+    clearAllFilters,
+    hasActiveFilters,
+  } = useCardFilters();
   const memberOptions = useMemo(
     () => getMembersFilterOptions(members),
     [members]
@@ -97,12 +102,6 @@ export function FilterMenu() {
       </li>
     );
   }
-
-  const hasActiveFilters = () => {
-    return (
-      filters.title?.trim() || filters?.members?.length || !!filters.noMembers
-    );
-  };
 
   function handleMembersAutocompleteChange(_, selected) {
     let newMembers = selected.map(m => m.id);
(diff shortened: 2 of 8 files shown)


src/assets/styles/components/FilterMenu.css
```diff
@@ -0,0 +1,14 @@
+.MuiAutocomplete-popupIndicator {
+  transform: none !important;
+  transition: none !important;
+}
```
snoxlax, line 4: it has to be `!important` ?
if disabling some functionalty of an MUI componnet is what you are trying to do they sometimes have a prop for that like
`disableAutoFocus` or `popupIcon={null}` or `disableAutoHighlight` etc

Edit: ive seen you already use some of them so there is no workaround?

src/components/FilterMenu.jsx
```diff
@@ -82,7 +138,76 @@ export function FilterMenu() {
...
+                  }
+                  indeterminate={(() => {
+                    const filtered = memberOptions.filter(
+                      m => m.id !== CURRENT_USER_ID_PLACEHOLDER
+                    );
+                    const selectedCount = filtered.filter(m =>
+                      filters.members?.includes(m.id)
+                    ).length;
+                    return selectedCount > 0 && selectedCount < filtered.length;
+                  })()}
```
snoxlax: Thats clever, but maybe we should just put it in a function outside? does it have to be inline?

src/components/FilterMenu.jsx
```diff
@@ -7,39 +8,95 @@ import TextField from "@mui/material/TextField";
...
 
-  const handleTitleChange = event => {
+  function handleTitleChange(event) {
     const value = event.target.value;
     setLocalTitle(value);
     updateFilterDebounced("title", value);
-  };
+  }
+
+  function handleSelectAllMembersChange(_, checked) {
```
snoxlax: is it possiable to combine this into one with handleMembersChange()?

snoxlax (approved):
Looks good very well done and clean!

my comments are just some questions and suggestions

sabellius:
[[FILTER-3] 🏷️ Implement Members Filtering](https://trello.com/c/pHOtSIqU/38-filter-3-%F0%9F%8F%B7%EF%B8%8F-implement-members-filtering)

src/assets/styles/components/FilterMenu.css
sabellius: There is no built in property to change this behavior, but I did gave it a more specific selector, and remove the `!important` to avoid overriding this element globally.

src/components/FilterMenu.jsx
sabellius: Done, it's a little longer than each individual one because we need to check what checkbox was toggled, but it does make since to combine them as most of the logic is similar.

src/components/FilterMenu.jsx
sabellius: Done, it's kinda long, so makes sense.

PR #25, opened 2025-10-24
feat: Created the ui effect of labels opening and closing #CATC-46

snoxlax:
## 🎯 Description

Implemented the open/close effect for labels across the board.

## 🔧 Changes

* Added a board-wide open/close state for labels.
* Prop drilled this state to cards so all labels update simultaneously.
* Updated label CSS to include a smooth open/close animation effect.

## 📝 Notes

Most of the changes are in the label CSS.

Open to suggestions on improving the cross-board effect currently it uses prop drilling, but we could consider using context instead. What do you think?

## 🖥️ Demo
[2025-10-24 21-27-10-1.webm](https://github.com/user-attachments/assets/b50a9156-3749-4467-93cb-5f86f3c9367e)


src/components/Card.jsx
```diff
@@ -27,6 +29,11 @@ export function Card({
     setAnchorEl(e.currentTarget);
   }
 
+  function handleClickLabels(e) {
+    e.stopPropagation();
+    setLabelsIsOpen(!labelsIsOpen);
```
sabellius: I'm still not sure when to use the function variation the the `setSomthing` of the `useState` hook, but from my general understanding, when your new state depends on the previous one, you should probably do this: `setLabelsIsOpen(prevValue => !prevValue);`

sabellius (changes requested):
Other than the single comment, looks very organized and clear what you did here. Good job!

src/pages/BoardDetails.jsx
```diff
@@ -67,6 +69,8 @@ export function BoardDetails() {
               <List
                 key={list.id}
                 list={list}
+                labelsIsOpen={labelsIsOpen}
```
LiorRey, line 100: Looks great! Nothing much to add - though right now (including my PR), we're already passing around 9 props to `List` 😅
Hopefully that’s not too much of an issue.

LiorRey (approved):

src/pages/BoardDetails.jsx
snoxlax: Yes I was saying the same maybe we should start using useContext for labels at least

src/components/Card.jsx
snoxlax: Yes, sometimes we should use `setLabelsIsOpen(prevValue => !prevValue);`, but in this case is it really matter because it’s just two states true or false and we’re basically flipping it either way

sabellius: (https://claude.ai/share/4010a31e-7107-4aca-ad4c-0155f097f8a1)

src/pages/BoardDetails.jsx
sabellius: I think we could just introduce a ui state slice to the store. Things like isSomethingOpen, activeListAction, activeListIndex, etc.. will solve us a lot of prop drilling. I just don't see a reason to use something like context when you already have a state management solution in place.

snoxlax: Yes or UI slice
But also the labels we have them every where it would be good to pass them from a store

PR #28, opened 2025-10-26
feature: Add Label Menu [CATC_59-add-label-menu]

LiorRey:
## 🎯 Description
Implemented a comprehensive label management system that allows users to create, edit, delete, and assign labels to cards. The feature includes a searchable label menu with real-time filtering and a visual label editor with color previews.

## 🔧 Changes
- **LabelMenu Component**: Main menu displaying all board labels with search functionality and label assignment checkboxes
- **LabelEditor Component**: Create/edit label interface with live color preview and 10 predefined color options
- **LabelMenuItem Component**: Individual label item with checkbox for card assignment and edit button
- **Label CRUD Operations**: Full create, read, update, delete functionality integrated with board state management
- **Card Label Assignment**: Toggle labels on/off for cards directly from the label menu
- **CSS Variables**: Defined label color scheme (bg, border, text) for 10 colors in `var.css`
- **Search & Filter**: Real-time label search with case-insensitive filtering

## 📝 Technical Notes
- Labels stored at board level; cards reference labels by ID
- New labels automatically assigned to current card upon creation
- Back navigation in list view closes menu; in editor view returns to list
- State properly resets when menu opens/closes to prevent stale data
- Fixed duplicate key warnings and card update logic to prevent duplicates

## 🎨 UI/UX Highlights
- Smooth view transitions between list and editor
- Live preview updates as user types label title
- Visual feedback for selected labels with checkboxes
- Responsive color grid with hover effects and selection indicators

### DIFF
--- a/src/components/CardModal.jsx
+++ b/src/components/CardModal.jsx
@@ -130,7 +130,6 @@ export function CardModal({
         </footer>
 
         <LabelMenu
-          key={anchorEl ? "open" : "closed"}
           boardId={boardId}
           listId={listId}
           card={card}
--- a/src/components/CardModal.jsx
+++ b/src/components/CardModal.jsx
@@ -13,6 +13,9 @@ export function CardModal({
   boardLabels = [],
   onClose,
   isOpen,
+  onSaveLabel,
+  onRemoveLabel,
+  onUpdateCard,
 }) {
   const [openSection, setOpenSection] = useState(false);
   const [anchorEl, setAnchorEl] = useState(null);
@@ -28,6 +31,24 @@ export function CardModal({
 
   function handleCommentSection() {
     setOpenSection(!openSection);
+  }
+
+  function handleToggleCardLabel(labelId) {
+    const currentLabelIds = card.labels || [];
+    let updatedLabelIds;
+
+    if (currentLabelIds.includes(labelId)) {
+      updatedLabelIds = currentLabelIds.filter(id => id !== labelId);
+    } else {
+      updatedLabelIds = [...currentLabelIds, labelId];
+    }
+
+    const updatedCard = {
+      ...card,
+      labels: updatedLabelIds,
+    };
+
+    onUpdateCard(updatedCard);
   }
 
   if (!card) return null;
@@ -110,9 +131,13 @@ export function CardModal({
 
         <LabelMenu
           boardLabels={boardLabels}
+          cardLabels={cardLabels}
           anchorEl={anchorEl}
           isLabelMenuOpen={isLabelMenuOpen}
           onCloseLabelMenu={onCloseLabelMenu}
+          onSaveLabel={onSaveLabel}
+          onRemoveLabel={onRemoveLabel}
+          onToggleCardLabel={handleToggleCardLabel}
         />
       </Box>
     </Modal>
--- a/src/components/CardModal.jsx
+++ b/src/components/CardModal.jsx
@@ -7,7 +7,9 @@ import AddIcon from "@mui/icons-material/Add";
 import { LabelMenu } from "./LabelMenu";
 
 export function CardModal({
-  list,
+  boardId,
+  listId,
+  listName,
   card,
   cardLabels = [],
   onEditCard,
@@ -43,7 +45,7 @@ export function CardModal({
     <Modal open={isOpen} onClose={onClose}>
       <Box className={`card-modal-box ${openSection ? "open" : "closed"}`}>
         <div className="card-modal-header">
-          {list.title}
+          {listName}
           <button className="icon-button" onClick={onClose}>
             <CloseIcon />
           </button>
@@ -128,12 +130,12 @@ export function CardModal({
         </footer>
 
         <LabelMenu
-          cardLabels={cardLabels}
+          boardId={boardId}
+          listId={listId}
+          card={card}
           anchorEl={anchorEl}
           isLabelMenuOpen={isLabelMenuOpen}
           onCloseLabelMenu={() => setAnchorEl(null)}
-          listId={list.id}
-          card={card}
         />
       </Box>
     </Modal>
--- a/src/components/CardModal.jsx
+++ b/src/components/CardModal.jsx
@@ -130,6 +130,7 @@ export function CardModal({
         </footer>
 
         <LabelMenu
+          key={anchorEl ? "open" : "closed"}
           boardId={boardId}
           listId={listId}
           card={card}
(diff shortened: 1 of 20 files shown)


src/components/LabelMenuItem.jsx
```diff
@@ -0,0 +1,45 @@
...
+      />
+      <label
+        htmlFor={`label-${label.id}`}
+        className="label-color-box"
+        style={{ backgroundColor: label.color }}
+      >
+        {label.title && <span className="label-title">{label.title}</span>}
+      </label>
+      <button className="label-edit-btn" aria-label="Edit label">
+        <svg
```
snoxlax: Why create a whole svg we have icons in MUI

src/components/LabelMenuItem.jsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <div className="label-menu-item">
+      <input
+        type="checkbox"
+        id={`label-${label.id}`}
+        checked={isSelected}
+        onChange={handleToggle}
+        className="label-checkbox"
+      />
+      <label
```
snoxlax, line 18: The <label> tag in HTML is used to associate text with a form input element like a text box, checkbox, or radio button.

src/components/LabelMenuItem.jsx
LiorRey: Through this svg, I made this exactly like in Trello :
image

src/components/LabelMenuItem.jsx
LiorRey: The label is indeed associated with the checkbox to its left :
image

When I press the label, it makes the checkbox checked, as if I checked the checkbox itself - exactly like in Trello.

src/components/LabelMenuItem.jsx
snoxlax: [this](https://fonts.google.com/icons?icon.set=Material+Icons&icon.query=pen) is from MUI
its ok if its not the same as Trello
also even if we use custom svg we wont hard code it into the code

src/components/List.jsx
```diff
@@ -41,9 +43,18 @@ export function List({
...
+      card.id === updatedCard.id ? updatedCard : card
+    );
+
+    setCards(updatedCards);
+
+    if (selectedCard && selectedCard.id === updatedCard.id) {
+      setSelectedCard(updatedCard);
+    }
+
+    const options = { key: "cards", value: updatedCards };
```
snoxlax: I dont understand what dose this do exactly? how is it affected by the labels menu?

src/components/LabelMenu.jsx
```diff
@@ -0,0 +1,139 @@
...
+      anchorEl={anchorEl}
+      isOpen={isLabelMenuOpen}
+      onClose={handleCloseMenu}
+      title={
+        view === "list"
+          ? "Labels"
+          : editingLabel
+          ? "Edit label"
+          : "Create label"
+      }
```
snoxlax: Why is it inside the props that are sent? and what dose this do?

src/components/List.jsx
LiorRey: The labels we choose for the card in LabelMenu, are also displayed in CardModal and in List:

As seen in List :
image

As seen in CardModal :
image



Also the old red code here was *adding* cards instead of updating them.
```
The code replaces the card instead of adding it :
const updatedCards = cards.map(card =>
  card.id === updatedCard.id ? updatedCard : card
);
```

src/components/LabelMenu.jsx
LiorRey: This determines what the title of the Popover will be :

1) When the Popover shows LabelMenu -> *title = "Labels"*
2) When the Popover shows LabelEditor, and we chose to edit an existing label -> *title = "Edit label"*
3) When the Popover shows LabelEditor, and we chose to create a new label -> *title = "Create label"*

LabelEditor  plays both roles of Add and Edit, same idea behind `CarEdit` that we saw in class.

Edit: I fixed by moving this `title` multi-lined code to a function: `getPopoverTitle()` .

src/components/LabelMenuItem.jsx
LiorRey: Fixed 🦾

src/store/actions/board-actions.js
snoxlax, line 1: Why this change?

src/pages/CardDetails.jsx
```diff
@@ -52,11 +52,11 @@ export function CardDetails() {
 
   return (
     <CardModal
-      cardLabels={cardLabels}
-      listTitle={list.name}
+      list={list}
```
snoxlax: Why do we need the whole list inside the Card Modal?

src/components/List.jsx
```diff
@@ -43,9 +43,19 @@ export function List({
...
+    );
+
+    setCards(updatedCards);
+
+    if (selectedCard && selectedCard.id === updatedCard.id) {
+      setSelectedCard(updatedCard);
+    }
+
+    const updates = { cards: updatedCards };
+    await onUpdateList(list, updates);
```
snoxlax: Please remove this we talked about it
Its a bad implementation of something else and its not related to Labels

src/pages/BoardDetails.jsx
snoxlax, line 1: Let’s try not to make these changes where we remove and then re-add the same line.

snoxlax (changes requested):
I’ve written down the things that need to be changed.

Also, I disagree with some of the changes made in CardModal.jsx, LabelEditor.jsx, and LabelMenu.jsx. I also need a clearer explanation of what’s being done inside them, since I don’t fully understand some parts.

src/components/CardModal.jsx
```diff
@@ -4,19 +4,24 @@ import CloseIcon from "@mui/icons-material/Close";
...
 import AttachFileIcon from "@mui/icons-material/AttachFile";
 import AddIcon from "@mui/icons-material/Add";
+import { LabelMenu } from "./LabelMenu";
 
 export function CardModal({
-  listTitle,
-  cardLabels = [],
+  boardId,
+  listId,
+  listName,
```
snoxlax: listTitle

LiorRey: At first I saw that there's no text in the CardModal header, and then I realized it's because I'm trying to get it from a field that does not exist in list: `list.title` .
I checked our data structure again, and I understood why there might be confusion around it :

image




So to be consistent, I chose `listName` over listTitle.

snoxlax (approved):
In good hour i approve well done!

PR #29, opened 2025-10-26
feature/CATC_54-implement-copy-list-in-list-actions

sabellius:
## 🎯 Description

This pull request introduces a copy menu form integrated into the list actions popover, which on submit, clones the desired list, inserts it after the original one, and updates the board object in the global state with the updated list of lists.

## 🔧 Changes

- 🟦 Created a reusable `ActionButton` component that wraps a MUI button.
- 📋 Added a `CopyListForm` that appears in the same popover as the list actions.
- 🛠️ Added board service functionality to clone a list and insert it after the original list.
- 🔄 Cloned the list on submit and updated the board object with the updated lists array.
- Refactor `boardService.copyList` and the `updateBoard` action along with `updateBoardWithActivity` to to be passed a `boardId` instead of a `board` object as an argument to prepare for future backend API that will probably be implement with an id reference.

## 📝 Notes
- The new components are designed for reusability and integration with existing UI patterns.
- Board state is updated immutably to ensure React state consistency.

## 🖥️ Demo

<details>
  <summary>Click to expand</summary>

  <p align="center">
    

https://github.com/user-attachments/assets/d79fd794-5dfb-478c-bc58-09fa689765c8

  </p>

</details>

### DIFF
--- a/src/components/CopyListForm.jsx
+++ b/src/components/CopyListForm.jsx
@@ -17,8 +17,9 @@ export default function CopyListForm({ initialValue, onCopy, onCancel }) {
 
   function handleSubmit(e) {
     e.preventDefault();
-    if (onCopy && textareaValue.trim()) {
-      onCopy(textareaValue.trim());
+    const newName = textareaValue.trim();
+    if (newName) {
+      onCopy(newName);
     }
     setTextareaValue("");
   }
@@ -45,7 +46,7 @@ export default function CopyListForm({ initialValue, onCopy, onCancel }) {
         className="copy-list-textfield"
         ref={textareaRef}
         value={textareaValue}
-        onChange={e => setTextareaValue(e.target.value)}
+        onChange={e => setTextareaValue(e.target.value.trim())}
         onKeyDown={handleKeyDown}
         minRows={2}
         autoFocus
--- a/src/components/CopyListForm.jsx
+++ b/src/components/CopyListForm.jsx
@@ -5,7 +5,7 @@ import Button from "@mui/material/Button";
 import Box from "@mui/material/Box";
 import ActionButton from "./ui/buttons/ActionButton";
 
-export default function CopyListForm({ initialValue, onCopy, onCancel }) {
+export default function CopyListForm({ initialValue, onSubmit, onCancel }) {
   const textareaRef = useRef(null);
   const [textareaValue, setTextareaValue] = useState(initialValue || "");
 
@@ -18,9 +18,9 @@ export default function CopyListForm({ initialValue, onCopy, onCancel }) {
   function handleSubmit(e) {
     e.preventDefault();
     const newName = textareaValue.trim();
-    if (newName) {
-      onCopy(newName);
-    }
+    if (!newName) return;
+
+    onSubmit(newName);
     setTextareaValue("");
   }
 
--- a/src/components/CopyListForm.jsx
+++ b/src/components/CopyListForm.jsx
@@ -46,7 +46,7 @@ export default function CopyListForm({ initialValue, onCopy, onCancel }) {
         className="copy-list-textfield"
         ref={textareaRef}
         value={textareaValue}
-        onChange={e => setTextareaValue(e.target.value.trim())}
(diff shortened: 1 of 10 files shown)


src/services/board/board-service-local.js
```diff
@@ -29,6 +30,40 @@ async function query() {
...
+  try {
+    const board = await getById(boardId);
+    const originalListIndex = board.lists.findIndex(l => l.id === listId);
+    if (originalListIndex === -1) throw new Error("List not found");
+
+    const listToCopy = board.lists[originalListIndex];
+
+    const clonedCards = listToCopy.cards.map(card => ({
+      ...card,
+      id: crypto.randomUUID(),
```
LiorRey, line 43: We are not using `makeId()` from `util-service.js`?
Is `randomUUID()` not creating really long IDs?

src/pages/BoardDetails.jsx
```diff
@@ -43,12 +43,29 @@ export function BoardDetails() {
...
+    }
+  }
+
   async function onRemoveList(listId) {}
 
   async function onUpdateList(list, { cardId = null, key, value }) {
     try {
       const options = { listId: list.id, cardId, key, value };
-      updateBoard(board, options);
+      updateBoard(board._id, options);
```
snoxlax, line 66: So we changed it to instead of sending the whole board we are now sending only the board._id?

src/components/CopyListForm.jsx
```diff
@@ -0,0 +1,74 @@
...
+        >
+          Name
+        </label>
+      </Box>
+      <TextareaAutosize
+        id="copy-list-textfield"
+        className="copy-list-textfield"
+        ref={textareaRef}
+        value={textareaValue}
+        onChange={e => setTextareaValue(e.target.value.trim())}
```
LiorRey: Doesn't this trim the value as you type, thus preventing users from entering spaces between words?

snoxlax (approved):
Looks good! 
just tell me about the updateboard fucntion if its changed

src/components/CopyListForm.jsx
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }, []);
+
+  function handleSubmit(e) {
+    e.preventDefault();
+    const newName = textareaValue.trim();
+    if (newName) {
+      onCopy(newName);
+    }
+    setTextareaValue("");
```
LiorRey, line 24: I think this can be inside the `if` block.

src/pages/BoardDetails.jsx
LiorRey: @snoxlax yes, personally I prefer not sending the whole object, but @sabellius can you please explain here shortly why it had to be changed eventually.

src/services/board/board-service-local.js
sabellius: I've discussed this with @snoxlax before. MongoDB generates similar UUID's to this (in terms of length) to ensure unique values for db entities. Since we're working locally and not with a real server we can use any kind of method for creating unique values. If this project wouldv'e been made from scratch, I wouldn't bother writing my own function for doing so like the `makeId()` in the starter. Either way, if you wish we use this function locally. I don't mind..

src/pages/BoardDetails.jsx
sabellius: @LiorRey @snoxlax If I understood the outcome of our discussion in Discord correctly, then - yes. My reasoning behind this change is that later we will have a server that will have endpoints for working with board object that will receive a boardId from the request json, and will perform a db query to find the board object using that boardId.

src/components/CopyListForm.jsx
sabellius: @LiorRey You are absolutely right! I now trim the value only on submit. Good catch!

src/components/CopyListForm.jsx
sabellius: I changed it. But also reversed the condition and returned early if the new title is empty.

LiorRey (approved):
Looks great. After our discussion, I approve 🦾

PR #30, opened 2025-10-31
(BEFORE THIS PR, WE NEED "CATC_59: LABEL MENU" PR) feature: Add drag and drop functionality for lists [CATC_20-add-drag-and-drop-functionality-for-lists]

LiorRey:
## 🎯 Description
Implements drag and drop functionality for reordering lists on the board. Users can now click and drag lists to reposition them, with real-time visual feedback showing the new order before dropping.

## 🔧 Changes
- Added drag and drop handlers (`onDragStart`, `onDragOver`, `onDragEnd`, `onDrop`) to List and BoardDetails components
- Implemented optimistic UI updates - lists stay in their new position immediately without waiting for backend confirmation
- Added visual feedback during drag (reduced opacity, grab cursor)
- Smart detection: no update or success message when dropping a list back to its original position
- Used `useRef` for `isUpdating` flag to avoid unnecessary re-renders

## 📝 Notes
- The drag and drop uses native HTML5 Drag and Drop API (no external libraries required)
- Lists reorder in real-time as you drag over other lists, providing instant visual feedback
- Backend updates happen asynchronously in the background
- If the backend update fails, the list order reverts with an error message

## 🧪 Testing
- Drag a list to a new position and verify it stays there
- Drag a list and drop it back in its original position - no update should occur
- Verify success/error messages display appropriately

### DIFF
--- a/src/assets/styles/components/List.css
+++ b/src/assets/styles/components/List.css
@@ -14,6 +14,24 @@
   color: var(--gray1);
 }
 
+.list-container:active {
+  cursor: grabbing;
+}
+
+.list-container.dragging {
+  opacity: 0.5;
+  cursor: grabbing;
+  box-shadow: 0 8px 16px rgba(0, 0, 0, 0.3);
+}
+
+.list-container:not(.dragging) {
+  opacity: 1;
+}
+
+.lists-list li {
+  transition: transform 0.2s ease;
+}
+
 .list-header {
   display: flex;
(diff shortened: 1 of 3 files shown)


snoxlax:
I’d suggest freezing this for now, since we’ll probably use a library. We can revisit and implement it after we figure out how to move lists and cards using the menus.

PR #34, opened 2025-11-02
Fixed the bug of the edit card from the card itself #CATC-69

snoxlax:
## 🎯 Description
This fixes the issue where editing a card from within the card itself was creating a duplicate instead of updating it.

## 🔧 Changes

* Fixed the bug by using the editCard action
* Moved the edit card into card
* Moved the delete card to the card
* Removed unnecessary code 

## 📝 Notes

This also effectively implements deleting a card directly from the card view, thanks to the simplified logic.

## 🖥️ Demo

<details>

  <p align="center">
  </p>

</details>

### DIFF
--- a/src/components/Card.jsx
+++ b/src/components/Card.jsx
@@ -7,18 +7,20 @@ import {
 import CardPopover from "./CardPopover";
 import { Box } from "@mui/material";
 import { useState } from "react";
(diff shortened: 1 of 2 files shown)


LiorRey (approved):
Indeed, it does look like an easy fix after that big change.
I approve 🦾

PR #35, opened 2025-11-03
feature/CATC_55-implement-move-all-cards-in-the-list-actions

sabellius:
## 🎯 Description

Implements "Move All Cards" feature for lists, allowing users to move all cards to existing lists or create a new "New List". Addresses ticket [CATC-55](https://trello.com/c/H7Qv1wUZ).

### Core Features

- ✨ Added "Move All Cards" menu option in list actions
- 🔧 Created `moveAllCards` and `createList` service functions
- 🔄 Added Redux actions and reducer cases for `ADD_LIST` and `MOVE_ALL_CARDS`

### Redux Infrastructure

- 🏗️ __Created Redux utilities__: `createAsyncActionTypes` (generates REQUEST/SUCCESS/FAILURE + auto-generated `.KEY`) and `createAsyncHandlers` (manages loading/error states)
- 🔄 __Refactored state management__: Dynamic loading/error keys instead of global flags (`loading.addList`, `errors.moveAllCards`)
- ✅ __Replaced string literals__ with constants for type safety
- 📝 __Added JSDoc documentation__ for utility functions

### Styling

- 🎨 Created reusable `.popover-menu` CSS class for consistent menu styling across the app

## 📝 Notes

- Supports moving to existing lists or creating "New List"
- Source list is emptied after move operation
- Current list disabled in destination menu to prevent self-move
- Granular state tracking enables multiple simultaneous async operations
- Established scalable pattern for future async actions

  ## 🖥️ Demo

  <details>

https://github.com/user-attachments/assets/aa0fd7fd-1920-40be-98f8-ae2f361c70f2

### AFTER CODE REVIEW CHANGES

https://github.com/user-attachments/assets/0d8ce8f5-e330-4307-9cb3-2bef42758713

  </details>

### DIFF
--- a/src/store/actions/board-actions.js
+++ b/src/store/actions/board-actions.js
@@ -117,6 +117,20 @@ export async function moveAllCardsToList(boardId, sourceListId, targetListId) {
   }
 }
 
+export async function createNewListAndMoveAllCards(boardId, sourceListId, listName = "New List") {
+  try {
+    const { updatedLists } = await boardService.createNewListAndMoveAllCards(
+      boardId,
+      sourceListId,
+      listName
+    );
+    store.dispatch(moveAllCardsAction(updatedLists));
+  } catch (error) {
+    store.dispatch(setError(`Error creating list and moving cards: ${error.message}`));
+    throw error;
+  }
+}
+
 export async function addCard(boardId, card, listId) {
   try {
     const newCard = await boardService.addCard(boardId, card, listId);
--- a/src/store/actions/board-actions.js
+++ b/src/store/actions/board-actions.js
@@ -21,25 +21,29 @@ import { boardService } from "../../services/board";
 
 export async function loadBoards() {
   try {
-    store.dispatch(setLoading(true));
+    store.dispatch(setLoading("loadBoards", true));
     const boards = await boardService.query();
     store.dispatch(setBoards(boards));
   } catch (error) {
-    store.dispatch(setError(`Error loading boards: ${error.message}`));
+    store.dispatch(
+      setError("loadBoards", `Error loading boards: ${error.message}`)
+    );
   } finally {
-    store.dispatch(setLoading(false));
+    store.dispatch(setLoading("loadBoards", false));
   }
 }
 
 export async function loadBoard(boardId, filterBy = {}) {
   try {
-    store.dispatch(setLoading(true));
+    store.dispatch(setLoading("loadBoard", true));
     const board = await boardService.getById(boardId, filterBy);
     store.dispatch(setBoard(board));
   } catch (error) {
-    store.dispatch(setError(`Error loading board: ${error.message}`));
+    store.dispatch(
+      setError("loadBoard", `Error loading board: ${error.message}`)
+    );
   } finally {
-    store.dispatch(setLoading(false));
+    store.dispatch(setLoading("loadBoard", false));
   }
 }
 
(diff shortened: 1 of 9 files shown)


src/store/actions/board-actions.js
```diff
@@ -90,6 +92,51 @@ export async function copyList(boardId, listId, newName) {
...
+      targetListId
+    );
+    store.dispatch(moveAllCardsAction(updatedLists));
+  } catch (error) {
+    store.dispatch(setError(`Error moving all cards: ${error.message}`));
+    throw error;
+  }
+}
+
+export async function createListAndMoveAllCards(
```
sabellius: @snoxlax I can try to implement an optimistic update - create a list with the cards already in it and display it immediately to avoid creating a separate action combining both creating a list and putting cards in it in one step. Let me know what you think

snoxlax (approved):
Looks good there’s quite a lot of code, but I understand most of it comes from the new store refactor you already explained.

Is there anything new here that we should pay extra attention to?

src/store/actions/board-actions.js
```diff
@@ -95,7 +142,7 @@ export async function addCard(boardId, card, listId) {
     const newCard = await boardService.addCard(boardId, card, listId);
     store.dispatch(addCardAction(newCard, listId));
   } catch (error) {
-    store.dispatch(setError(`Error adding card: ${error.message}`));
+    store.dispatch(setError("addCard", `Error adding card: ${error.message}`));
```
LiorRey, line 145: If we intend to use the string `"addCard"`  again anywhere else under the context of Error, I prefer it to be stored in a constant somewhere.
The same goes for the other strings here in the service, like: `"editCard"`, `"deleteCard"`, `"moveList"`, etc.

LiorRey (approved):
I see that a lot of the code is regarding the new approach we discussed on Discord.
Other than my remark (about putting the error-related strings in constants if we use those strings again) - I approve 🦾

src/store/actions/board-actions.js
sabellius: I absolutely agree, what you suggested is one of the solutions I had in mind:
```js
export const LOADING_KEYS = {
  // Regular async functions
  LOAD_BOARDS: "loadBoards",
  LOAD_BOARD: "loadBoard",
  CREATE_BOARD: "createBoard",
  UPDATE_BOARD: "updateBoard",
  DELETE_BOARD: "deleteBoard",
  COPY_LIST: "copyList",
  ADD_CARD: "addCard",
  EDIT_CARD: "editCard",
  DELETE_CARD: "deleteCard",
  MOVE_LIST: "moveList",
  
  // Async action types (matches createAsyncHandlers usage)
  ADD_LIST: "addList",
  MOVE_ALL_CARDS: "moveAllCards",
};
```

However, this seems very repetitive as well, so I add a dynamic creation of this key string inside `createAsyncActionTypes` which we will use almost all of them time when dealing with async operations. If for some reason we need error/loading state for a non-async action or some other special cases we can create such an array only for those actions - but until we get to see such scenario I think we should use this in the following manner:
```js
const SOME_ASYNC_OPERATION = createAsyncActionTypes("SOME_ASYNC_OPERATION")
// Returns:
// {
//   REQUEST: "SOME_ASYNC_OPERATION_REQUEST",
//   SUCCESS: "SOME_ASYNC_OPERATION_SUCCESS",
//   FAILURE: "SOME_ASYNC_OPERATION_FAILURE",
//   KEY: "someAsyncOperation"
// }

// then use with setError/Loading like this:
setLoading(SOME_ASYNC_OPERATION.KEY, true/false);
```

For the existing actions, i used a simple string value for now until we refactor the whole file. I used the new helper function on the actions I worked on in this branch to showcase the difference between how we do things at the moment vs how better it can get.

NOTE: Please notice that when dealing with async operations, you most likely won't need to use `setError/Loading` manually as the `createAsyncHandlers` will handle changes to those states automatically unless some customization will be needed.

PR #39, opened 2025-11-07
hotfix/CATC_84-fix-eslint-configuration

sabellius:
## 🎯 Description
Fixes ESLint configuration errors that prevented the linter from working correctly. The config was using the old eslintrc format which is incompatible with ESLint v9's flat config system.

## 🔧 Changes
- 🔄 Migrated ESLint config from eslintrc format to flat config format
- 🔌 Converted `plugins` from array format to object format (flat config requirement)
- ⚛️ Added `eslint-plugin-react` dependency and import
- ⚙️ Added React version auto-detection in settings (`version: 'detect'`)
- 📦 Moved `js.configs.recommended` to top-level array position
- 🔧 Spread plugin rules directly into rules object instead of using `extends`

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,18 +1,20 @@
 import js from '@eslint/js'
 import globals from 'globals'
+import react from 'eslint-plugin-react'
 import reactHooks from 'eslint-plugin-react-hooks'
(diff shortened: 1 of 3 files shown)


snoxlax (approved):
Looks good!

PR #41, opened 2025-11-07
Bugfix/catc 66 fix title property naming inconsistency

sabellius:
## 🎯 Description
This PR fixes naming inconsistencies in the board and list entities. 

## 🔧 Changes
- The `name` property in lists and boards were changed to `title` project wide.

## 📝 Notes
- After this is merged, we need to re-create the boards data locally. As I didn't provide a migration plan.


snoxlax (approved):
Great work looks good!

PR #43, opened 2025-11-10
feature/CATC_64-allow-users-to-signup

sabellius:
## 🎯 Description
This feature intends to add functionality that enables users to sign up to the app.

## 🔧 Changes
- Add a local and remote `auth` services to separate authentication logic from users related logic.
- Add an `auth` reducer an action for managing `currentUser` and other authentication related state.
- Add Signup page with a form based on Trellos' design.

### Design
Trello's original signup workflow has multiple steps and some additional components that not necessarily should be implemented in the same way in our app so I made some modifications.
<details>
  <summary>Click to expand</summary>

  <p align="center">
    <h3>Trello's signup flow</h3>
Screenshot from 2025-11-08 19-59-14

 <h3>Our signup flow</h3>

  </p>

</details>

#### Modifications
- Avoid doing email verification.
- Use a single form for all the relevant data.
- Remove the signup using external services (Google, Slack, etc..).
- Remove some company related texts.
- Keep the marketing checkbox and legal section without company specific details for future implementation.

#### Technical notes
- The user is automatically logged in after signing up to avoid friction.
- The user is saved in session storage to mimic some sort of server authentication memory (without the password for obvious reasons).
- Avoid nesting `signup/login` routes under `auth` because it's not needed in our implementation.
- I grouped the variables I added in this feature under the components file instead the `vars.css` because we haven't yet figured out a design system that uses those variables in a consistent manner. so I've put them in this separate file not to pollute the global variables and for easier future refactoring.

## 🖥️ Demo

<details>
  <summary>Watch Demo Video</summary>

  <p align="center">

https://github.com/user-attachments/assets/dfa3e951-93c4-4e48-88f8-465d0e52729c

  </p>

</details>

PR #44, opened 2025-11-11
feature/CATC_63-allow-users-to-login

sabellius:
## 🎯 Description
This feature intends to add functionality that enables users to login to the app. This PR is based on https://github.com/LiorRey/ca-final-project/pull/43 which contains the signup form which was refactored here.

## 🔧 Changes
- Add a login page with a form based on Trellos' design.
- Refactor the extracted duplicate layout logic from the signup form to a separate component to be reused in the login form.
- Modify the user data generator to create an email property for users.
- Use a function inside the board data generator to create boards with a test user with predefined email and password for easy access.

### Design
Trello's original login flow has multiple steps and some additional components that not necessarily should be implemented in the same way in our app so I made some modifications.
<details>
  <summary>Click to expand</summary>

  <p align="center">
    <h3>Trello's login flow</h3>

 <h3>Our signup flow</h3>

  </p>

</details>

#### Modifications
- Remove the signup using external services (Google, Slack, etc..).
- Remove some company related texts.
- Keep the "Remember me" checkbox for future implementation.

## 🖥️ Demo

<details>

  <p align="center">

https://github.com/user-attachments/assets/e75b9502-7504-4d32-a433-94ee4dcbbf2b

  </p>

</details>


src/components/SignupForm.jsx
```diff
@@ -115,197 +114,175 @@ export function SignupForm({ onSubmit }) {
...
+                fullWidth
+                variant="outlined"
+                className="auth-text-field"
+                error={!!errors.fullName}
+                helperText={errors.fullName?.message}
               />
-            </Box>
+            )}
+          />
+        </Box>
```
snoxlax: If possible, maybe this one could also be a custom auth form field?

src/assets/styles/components/auth/AuthFormLayout.css
snoxlax, line 1: Maybe we could use the [MUI Theming](https://mui.com/material-ui/customization/theming/)?
It could help us write less CSS code.

src/components/LoginForm.jsx
```diff
@@ -0,0 +1,151 @@
...
+                          {showPassword ? <VisibilityOff /> : <Visibility />}
+                        </IconButton>
+                      </InputAdornment>
+                    ),
+                  },
+                }}
+              />
+            )}
+          />
+        </Box>
```
snoxlax: Another example maybe we could put this in a custom component if possible

snoxlax (changes requested):
Looks good! I also added some comments and suggestions. 
I think the validations should be in a separate place from the component 
it would be more organized and easier to update in the future.

src/components/SignupForm.jsx
sabellius: Are you talking about `fullName` specifically, or the form input in general? In what other place in the app will we need another `fullName` field?

src/assets/styles/components/auth/AuthFormLayout.css
sabellius: Can you elaborate? Do you mean using this theme functionality for the entire app or just for this feature? It's seems implementing this behavior (the one mentioned in the link) means abandoning  entirely/partialy the design system we currently have based on css files (at least at the setup/vars/general stuff kind of level).

src/components/LoginForm.jsx
sabellius: Why though? Where else in the app we need or would need to implement the same exact password strength logic? The only other input that is a password input in the app is in the login form, but it doesn't validate strength nor anything else, just a regular controlled input with a regular `onChange` handler. Why would we want to extract this logic into a separate component although it's not re-used anywhere else (nor will it be, most likely)?

snoxlax (approved):
Looks good and very clean!

PR #47, opened 2025-11-16
fix: Remove deleted labels from all cards in board

LiorRey:
## 🎯 Description

Fixed issue where deleted labels remained as IDs in cards across the board.
Now, deleted labels are completely removed from all cards - both visually and from their data.

## 🔧 Changes

- Update `deleteLabel` in `boardService` to remove `labelId` from all cards
- Update `DELETE_LABEL.SUCCESS` reducer to clean cards' `labels` array
- Simplify `LabelMenu.handleDeleteLabel` (no manual toggle needed)
- Ensures deleted labels are removed from the entire board, not just from the opened card

### DIFF
--- a/src/components/LabelMenu.jsx
+++ b/src/components/LabelMenu.jsx
@@ -58,11 +58,6 @@ export function LabelMenu({
 
   async function handleDeleteLabel(labelId) {
     await deleteLabel(boardId, labelId);
-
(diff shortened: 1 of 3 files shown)


sabellius (approved):
Looks just like we discussed. Good job! 💪

PR #48, opened 2025-11-21
Feature/catc 91 implement fractional positioning in lists

sabellius:
## 🎯 Description

Implemented __fractional indexing__ for list positioning to replace array index-based ordering. Lists now have explicit `position` properties (e.g., `"a0"`, `"a1"`, `"Zy"`) that enable efficient reordering without array manipulation or re-indexing.

This change prepares the codebase for MongoDB migration and improves performance by updating only the moved list's position field instead of splicing arrays.

---

### ✨ New Features

- __Added fractional indexing service__ (`fractional-index-service.js`)

  - Utilities for generating and managing fractional positions
  - Uses `fractional-indexing` npm package
  - Case-sensitive native string comparison for sorting (per package requirements)

### 🔨 Service Updates

- __Updated `createList()`__ - Automatically generates positions for new lists
- __Updated `copyList()`__ - Calculates position for copied lists relative to original
- __Updated `moveList()`__ - Now updates position field instead of array splicing
- __Updated `getById()`__ - Sorts lists by position on load

### 🗃️ Data Layer

- __Updated data generator__ - Creates initial fractional positions for all lists
- __Updated `getEmptyList()`__ - Includes `position: null` (set when added to board)

### ⚠️ Breaking Changes

- __Requires clearing localStorage__ - Existing boards without positions will not work
- All lists must have a `position` property (no migration support)

### 🚀 Benefits

- __Efficient moves__ - Update one property instead of splicing arrays
- __No re-indexing__ - Other lists aren't affected when moving a list
- __Backend-ready__ - MongoDB-friendly (update single document field)
- __Infinite positions__ - Can always insert between any two lists

### DIFF
--- a/frontend/src/services/board/board-service-local.js
+++ b/frontend/src/services/board/board-service-local.js
@@ -3,6 +3,11 @@ import { loadFromStorage, saveToStorage } from "../util-service";
 import boardDataGenerator from "./board-data-generator";
 import { USER_STORAGE_KEY } from "../user/user-service-local.js";
 import { getFilteredBoard } from "../filter-service";
+import {
+  calculateNewPosition,
+  generatePositionAtEnd,
+  sortByPosition,
+} from "./fractional-index-service";
 
 const BOARDS_STORAGE_KEY = "boardDB";
 _createBoards();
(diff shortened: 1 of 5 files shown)


frontend/src/services/board/board-service-local.js
```diff
@@ -169,7 +179,12 @@ export async function moveList(
 
 async function getById(boardId, filterBy = {}) {
   try {
-    const board = await storageService.get(BOARDS_STORAGE_KEY, boardId);
+    let board = await storageService.get(BOARDS_STORAGE_KEY, boardId);
```
snoxlax: this can be const

snoxlax (approved):
Looks good but since this function handles position management for lists, and we also need similar logic for cards, maybe we can generalize it so it works for both.

frontend/src/services/board/board-service-local.js
sabellius: Done

snoxlax (approved):
Looks good!

LiorRey (approved):
After reviewing and discussing positioning in Discord, I approve 🦾

PR #52, opened 2025-11-24
Cleanup and reorder of files and code #CATC-98

snoxlax:
## 🎯 Description

Major cleanup and reordering of files and code across the project. This PR removes redundant code, restructures services, and fixes several minor bugs.

## 🔧 Changes

* Reordered and cleaned up large parts of the codebase
* Removed unnecessary and unused code/files that we will not be using
* Reorganized imports and improved overall file structure
* Updated and refactored several service functions
* Reworked the **copy** function to match the conventions we agreed upon (it previously relied on `updateBoard`)
* Fixed multiple small bugs and inconsistencies
* Removed old **car** and **reviews** related files, including their CSS
* Reordered logic inside the board service
* Addressed several errors (some still remain as discussed, we’ll handle them together)
* **There may still be additional inconsistencies or cleanup opportunities, but this PR is already large, so I avoided digging even deeper into the code for now**

## 📝 Notes

Some errors are still present, specifically the ones we discussed earlier. We'll resolve those together in the next step.
There’s also likely more to improve, but this PR already covers a substantial amount of cleanup.

## 🖥️ Demo

<details>

  <p align="center">
  </p>

</details>


frontend/src/components/CardModal.jsx
```diff
@@ -101,8 +101,9 @@ export function CardModal({
                 className="description-input"
                 placeholder="Add a description"
                 spellCheck="false"
-                value={card.description}
-                onChange={e => setDescription(e.target.value)}
+                value={cardDetails.description}
+                onChange={e => handleChangeCard("description", e.target.value)}
+                onBlur={handleSaveCard}
```
sabellius, line 106: Do we really want to save it `onBlur`? In Trello you need to hit the "Save" button to actually save the card.

frontend/src/components/Header.jsx
```diff
@@ -14,7 +14,7 @@ export function Header() {
       navigate("/");
       showSuccessMsg(`Bye now`);
     } catch (err) {
-      showErrorMsg("Cannot logout");
+      showErrorMsg("Cannot logout: " + err.message);
```
sabellius: I think we actually show get rid of all the `showErrorMsg` and `showSuccessMsg` becuase Trello doesn't show these kind of alerts and instead it only shows the following popover after certain actions in the app:
Image

frontend/src/pages/Chat.jsx
```diff
@@ -18,6 +18,10 @@ export function ChatApp() {
 
   const botTimeoutRef = useRef();
 
+  function addMsg(newMsg) {
```
sabellius: Do we even need this component? @LiorRey @snoxlax

sabellius (changes requested):
Please see my comments. Besides those, looks great!

frontend/src/components/CardModal.jsx
snoxlax: Good point. I'll remove it for now, and when I get to the description, I'll implement it correctly.

frontend/src/pages/Chat.jsx
snoxlax: No, we do not it will be removed
If we ever decide to add something like this, we would likely implement it ourselves using MUI.

sabellius (approved):

PR #53, opened 2025-11-26
List model with service and controller #CATC-101

snoxlax:
## 🎯 Description

Created full **routes, controller, and service layers** for the **List** model, implementing all CRUDL functionality and related operations.
Also added a centralized **config system (`env.js`)** to load and manage environment variables from a single place, making them easier to access, maintain, and refactor.

## 🔧 Changes

* Added full List **routes**
* Implemented List **controller** with all endpoints (CRUDL + extra actions)
* Added List **service** layer for business logic
* Created **env.js** for centralized environment variable handling
* Improved project structure and separation of concerns

## 📝 Notes

* Environment variables can now be imported cleanly from one file (`env.js`)
* New List API endpoints follow REST conventions and are fully integrated
* Further improvements (refactoring, validation, error handling) can be added later if needed

## 🖥️ Demo

<details>

  <p align="center">
  </p>

</details>


backend/src/config/env.js
```diff
@@ -0,0 +1,27 @@
...
+  const env = process.env.NODE_ENV;
+
+  const port = parseInt(process.env.PORT || "3000", 10);
+
+  const dbURL = process.env.MONGODB_URI || "mongodb://localhost:27017/basicdb";
+
+  const dbName = process.env.MONGO_DATABASE || "basicdb";
+
+  const frontendUrl = process.env.CORS_ORIGIN || "http://localhost:5173";
+
```
sabellius, line 11: when not just use the values directly in the returned object? Each variable is only used once so I'm not sure we need them in the first place. What do you think? Also, I don't we should resort to development only fallback, if this file's purpose is to handle loading the correct environment variables, it only make sense to have fallback values that we don't mind being undefined in the first place.

backend/src/config/env.js
```diff
@@ -0,0 +1,27 @@
+export function getConfig() {
+  const env = process.env.NODE_ENV;
+
+  const port = parseInt(process.env.PORT || "3000", 10);
```
sabellius: I for some reason thought `dotenv` knows how to differentiate between different data types (or at least string and numbers). Perhaps we could load our config from a [YAML](https://learnxinyminutes.com/yaml/) / [TOML](https://learnxinyminutes.com/toml/), [JSON](https://learnxinyminutes.com/json/) file? This will allow us to use different data types other than string of variable values (even arrays) and some of them even support more advanced features. What do you guys think? @snoxlax @LiorRey 

Regardless, same comment on the fallback here, and also, `parseInt` by default uses base 10 for parsing so there is no need to explicitly pass it.

backend/src/controllers/list-controller.js
```diff
@@ -0,0 +1,89 @@
+import * as listService from "../services/list-service.js";
+import createHttpError from "http-errors";
```
sabellius: It looks like an external package, but I can find a `package.json` in this PR. Is this by mistake?

backend/src/services/list-service.js
```diff
@@ -0,0 +1,59 @@
...
+  return lists;
+}
+
+export async function updateList(id, updateData) {
+  const list = await List.findByIdAndUpdate(id, updateData, {
+    new: true,
+    runValidators: true,
+  });
+  return list;
+}
```
sabellius, line 21: Same here, if the variable is not used elsewhere in the function and is only returned - why is it necessary in the first place?

backend/src/services/list-service.js
```diff
@@ -0,0 +1,59 @@
...
+
+export async function repositionList(id, position) {
+  const list = await List.findById(id);
+  if (!list) {
+    return null;
+  }
+  list.position = position;
+  await list.save();
+  return list;
```
sabellius: Mongoose has a vast api and one of the function in it is [`Model.findByIdAndUpdate()`](https://mongoosejs.com/docs/api/model.html#Model.findByIdAndUpdate()) which you already used in the code elsewhere in the file, and does exactly what you're trying to achieve here in a single line..
```js
export async function repositionList(id, position) {
  return await List.findByIdAndUpdate(id, { position }, { new: true });
}
```
If it's found it returns the updated list, otherwise returns `null`.

Same goes with `export async function archiveList(id)`

backend/src/services/list-service.js
```diff
@@ -0,0 +1,59 @@
...
+
+export async function updateList(id, updateData) {
+  const list = await List.findByIdAndUpdate(id, updateData, {
+    new: true,
+    runValidators: true,
+  });
+  return list;
+}
+
+export async function repositionList(id, position) {
```
sabellius: Why do we now use `reposition` over `move`? Do we prefer one over the other? @snoxlax @LiorRey 
I'm down to pick one and stick with it in the entire app instead of using multiple terms to describe the same action/resource.

backend/src/models/List.js
```diff
@@ -0,0 +1,36 @@
...
+  boardId: {
+    type: String,
+    required: true,
+  },
+  title: {
+    type: String,
+    required: true,
+    trim: true,
+  },
+  description: String,
```
sabellius: Won't it be better to use a default value for this?

backend/src/models/List.js
```diff
@@ -0,0 +1,36 @@
...
+  position: {
+    type: String,
+    required: true,
```
sabellius: Is it required? How about archived lists? do they need a `position`?

backend/src/models/List.js
```diff
@@ -0,0 +1,36 @@
...
+    default: null,
+  },
+  createdAt: {
+    type: Number,
+    default: Date.now,
+  },
+  updatedAt: {
+    type: Number,
+    default: Date.now,
+  },
```
sabellius: Why are those numbers? Mongoose has a `Date` schema type which is the correct type to use here.
Also, when defining a schema, since adding `createdAt` and `updatedAt` to object is a common practice, mongoose supports adding them using the [`timestamps`](https://mongoosejs.com/docs/timestamps.html) option, which in my opinion should be always used instead of defining them manually.

backend/src/routes/index.js
```diff
@@ -1,9 +1,11 @@
 import express from "express";
 
 import authRoutes from "./auth.js";
+import listRoutes from "./list.js";
 
 const router = express.Router();
 
 router.use("/api/auth", authRoutes);
+router.use("/api/list", listRoutes);
```
sabellius: In [`REST`](https://www.restapitutorial.com/introduction/restquicktips), the naming convention for resources is to use the plural form except for cases when there is no sense in doing so like in cases like `auth`.[Another reference](https://www.restapitutorial.com/introduction/resourcenaming)

backend/src/routes/list.js
```diff
@@ -0,0 +1,22 @@
...
+  getListByBoardId,
+  updateList,
+  repositionList,
+  archiveList,
+  deleteList,
+} from "../controllers/list-controller.js";
+
+const router = express.Router();
+
+router.get("/by-board/:boardId", getListByBoardId);
```
sabellius: This is another anti-pattern in `REST`. The more appropriate way of doing this is creating a nested route such as `/api/boards/:boardId/lists/:listId/cards` (or just `/api/boards/:boardId/cards` if you want all the cards in the board like in this example). 
There is no point creating a special route for filtering cards by board, because using 
```js
router.get("/", getAllCards);
```
and sending a `boardId` among other filters is in my opinion the correct way to do this.

Currently' it seems there is no such route, so all card filtering is non-existent?

backend/src/services/list-service.js
```diff
@@ -0,0 +1,59 @@
+import { List } from "../models/List.js";
+
+export async function createList(listData) {
+  const list = new List({
+    ...listData,
+    createdAt: Date.now(),
+    updatedAt: Date.now(),
+    archivedAt: null,
+    deletedAt: null,
```
sabellius: https://github.com/LiorRey/ca-final-project/pull/53/files#diff-5f47d0c50c924ab7d512eb55c088d61a67d0f773523f404c40a64a004994d980R18-R24
These should be created with default value from the model and not manually..

backend/src/controllers/list-controller.js
```diff
@@ -0,0 +1,89 @@
...
+
+export async function createList(req, res, next) {
+  try {
+    const list = await listService.createList(req.body);
+    res.status(201).json({ success: true, data: list });
```
sabellius: Why do we need `success`? Isn't that data is not `null` enough? @LiorRey @snoxlax

backend/src/controllers/list-controller.js
```diff
@@ -0,0 +1,89 @@
...
+  } catch (error) {
+    next(error);
```
sabellius: Where does the error go?

backend/src/controllers/list-controller.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+
+export async function deleteList(req, res, next) {
+  try {
+    const list = await listService.deleteList(req.params.id);
+    if (!list) {
+      throw createHttpError(404, "List not found");
+    }
+    res.json({ success: true, data: list });
```
sabellius: Why do we need the list if we deleted it?

backend/src/controllers/list-controller.js
```diff
@@ -0,0 +1,89 @@
...
+export async function archiveList(req, res, next) {
+  try {
+    const result = await listService.archiveList(req.params.id);
+    if (result.error === "not_found") {
+      throw createHttpError(404, "List not found and was not archived");
+    }
+    if (result.error === "already_archived") {
+      throw createHttpError(400, "List is already archived");
+    }
+    res.json({ success: true, data: result.data });
```
sabellius: Explain the flow here please, and why does it seems different than `deleteList`.

sabellius (changes requested):
See my comments

backend/src/controllers/list-controller.js
snoxlax: It comes with express as one of the dependencies

backend/src/models/List.js
snoxlax: I mean if its unarchived it can return with its previous position i think
I think a list should always be postioned somewhere no?

backend/src/controllers/list-controller.js
snoxlax: I dont know what about when in other cases where you can retrun a message which is not an error but its not the data you are looking for and you can know that by seeing `success : false` 
we can get rid of it if you want

backend/src/controllers/list-controller.js
snoxlax: Here we only update the archivedAt field with the date it was archived, and in the delete operation we permanently remove the list

backend/src/controllers/list-controller.js
snoxlax: We return the list so the UI can update with the information of the list that was deleted. Don’t we need that? We can also just return an “OK” response or even return nothing.

We may also need it for activity logging purposes.

backend/src/routes/list.js
snoxlax: Its for lists its to get all lists that are related to that board id 
will fix it to this 

```js 
router.get("/", getAllLists);
```

yes there is no filtering yet its still just basic crud operations with some extra things we might want to change

backend/src/controllers/list-controller.js
snoxlax: This was intended for the error handler we agreed to implement, but since we still need to determine the approach, it’s temporary for now.

backend/src/controllers/list-controller.js
sabellius: Isn't the presence of data means the result of the request is success? 1if (success) === if (!!data)`?
I just don't want request to return a resoponse in one format and other return it in a different one.. thats gonna make the services in the frontend a mess.

backend/src/controllers/list-controller.js
sabellius: When this functionality will be implemented, and it's purpose is to avoid the `try..catch` in the controllers we will need to modify this controller action either way so we should for now work we the code as if it possible wont exist.

backend/src/controllers/list-controller.js
sabellius: I think that we can agree that as a rule of thumb, keeping things in the code because of some future functionalities may or may not be implemented is a bad idea? Do you already knowhow activities would be implemented? are you sure its going to use this list? If you need the id of the list to know which list to remove in the front, you can just pass the id back to the front, otherwise, there is not need to return any value are that true of false in my opinion.

backend/src/models/List.js
sabellius: It depends on how the archived lists bin and the un-archive action are going to be implemented. But it's ok to keep it this way for now.

sabellius (approved):

PR #56, opened 2025-11-27
Fixed the card menu visually and removed the dependency on context

snoxlax:
## 🎯 Description

Refactored the card menu to better match the original design and replaced the previous `Select` implementation with a more suitable component. This also removed the need for `useContext` and unnecessary state. Updated MUI theme usage to style components consistently and removed old CSS that became redundant after the theme was introduced.

## 🔧 Changes

* Reworked the card menu UI to improve alignment and match the intended design
* Replaced `Select` with a more appropriate component
* Removed `useContext` and extra state previously required
* Applied MUI theme overrides to handle styling instead of custom CSS
* Cleaned up leftover CSS related to the old implementation
* Fixed card alignment issues

## 📝 Notes

This isn’t a new feature—it's a cleanup and improvement of existing UI elements. The result is better visuals, improved consistency, and less code to maintain long-term.

## 🖥️ Demo

<details>

  <p align="center">
  </p>

https://github.com/user-attachments/assets/5cf3dd64-ab13-4bbe-bf44-0ac4184fdf29

</details>


frontend/src/assets/styles/setup/var.css
```diff
@@ -5,6 +5,7 @@
   --clr4: #9cbed6;
   --clr5: #f0f5f3;
   --clr6: #00416f;
+  --primary-color: #579dff;
```
sabellius: Same argument I've told to @LiorRey - perhaps we could benefit from using the existing `atlaskit` vars.

frontend/src/components/card/CardActionForm.jsx
```diff
@@ -24,23 +23,22 @@ export function CardActionForm({
...
+      currentList?.cards?.findIndex(c => c.id === card?.id) ?? 0;
+
     const baseValues = {
-      boardId: board?._id || "",
-      listId: board?.lists?.[0]?.id || "",
-      position: 0,
+      boardId: board._id || boards[0]._id || "",
+      listId: listId || board?.lists?.[0]?.id || "",
+      position: currentPosition,
     };
```
sabellius, line 57: I was thinking, and perhaps we should get this data from the store at all. we could just add an endpoints for getting this minimal data in the relevant controllers in the server and once a user clicks on this, we get an up-to-date data from the server rather than relying on this data from the store which becomes stale after we perform this action (we we then need to update in the store for it to reflect the new data in the ui select elements). I think a better solution for this will be to mock such endoint in the local service - something like this:
```js
// get options for boards
export async function getBoardPreviews() {
  // get an array of board names, ids, etc..
}

export async function getListPreviews(boardId) {
  // get an array of list names, ids, etc.. (of a specific board)
}

// or perhaps add a single endpoint that does both?
```

frontend/src/components/ui/CustomAutoComplete.jsx
```diff
@@ -0,0 +1,100 @@
...
+      <li
+        key={option.id}
+        {...rest}
+        style={{
+          ...rest.style,
+          borderLeft: isSelected
+            ? "3px solid #1976d2"
+            : "3px solid transparent",
+        }}
+      >
```
sabellius: I think we should avoid inline style as much as possible, as it is hard to overwrite in terms of specificity.

frontend/src/theme/muiTheme.js
```diff
@@ -241,6 +241,112 @@ const muiTheme = createTheme({
       )
     ), // Indices 3-24 (22 elements)
   ],
+
```
sabellius, line 248: I think the whole point of this file is to be used with `atlaskit` given variables for as much of the app as possible without relying on our own random css vars (such as `gray2`), as well as some magic values such as `"#e0e0e0"` or `rgba(87, 157, 255, 0.3)"`

sabellius (changes requested):
Please see my comments.

frontend/src/components/card/CardActionForm.jsx
snoxlax: Yeah thats probably a better approach
So you’re suggesting that instead of pulling the data from the store, the component should call these functions on the local service directly (which will then fetch from the backend) and receive the updated data straight from the source?

snoxlax: I updated it, but the board service still needs refinement. Right now it only has the function signatures it returns the values we need, but the actual implementation will probably change later.

sabellius (approved):

PR #59, opened 2025-11-28
feature: Add Card from List Actions Menu [CATC_28-add-card-from-list-actions-menu]

LiorRey:
## 🎯 Description
Implemented the option from "List Actions" menu to add a card to the top of the list.

## 🔧 Changes
- Extracted the specific form with its logic of adding new cards to a list, to a new AddCardForm component.
- AddCardForm can only appear in one list on the board at a given time.
- AddCardForm can appear only once inside a list at a given time - at its top or at its bottom.

## 📝 Technical Notes
- When AddCardForm is in the top of the list, cards will be added in a stack-like behaviour, meaning the newest card to be added at the top of the list will push down all other cards.

## 🎨 UI/UX Highlights
- Upon clicking on "List Actions" -> "Add card" option - a new header containing AddCardForm will be shown at the top of the list.
- Enabled automatic scrolling to the top of the list when AddCardForm is shown at the top of the list, or when a card is added at top.


frontend/src/assets/styles/components/AddCardForm.css
```diff
@@ -0,0 +1,48 @@
...
+  display: flex;
+  align-items: center;
+  gap: 5px;
+}
+
+.add-card-buttons-container .add-card-contained-button {
+  width: fit-content;
+  padding: 0.4rem 0.75rem 0.3rem 0.75rem;
+  color: var(--button-black-text);
+  background-color: var(--button-soft-blue-bg);
```
sabellius: I think we should stop using the color from variables that we created and use the colors from the `atlaskit` package? You can see what variable Trello used in their app, and try an use the variable of the same name in our app. In my opinion it is a much better solution than reusing inconsistent variables with inconsistent names or colors. What do you guys think? @LiorRey @snoxlax  

Image

frontend/src/components/List.jsx
```diff
@@ -54,40 +54,42 @@ export function List({
...
-      scrollListToEnd({ direction: SCROLL_DIRECTION.VERTICAL })
-    );
+    requestAnimationFrame(() => {
+      requestAnimationFrame(() =>
+        scrollListToEdge({
+          direction: SCROLL_DIRECTION.VERTICAL,
+          scrollToBottom: addCardToEnd,
+        })
+      );
+    });
```
sabellius, line 71: Why is there a nested call to `requestAnimationFrame`?

frontend/src/components/List.jsx
```diff
@@ -54,40 +54,42 @@ export function List({
...
+      : [addedCard, ...list.cards];
 
-    const newCard = {
-      ...boardService.getEmptyCard(),
-      title: newCardTitle,
-      createdAt: Date.now(),
-    };
-
-    const updatedCards = [...cards, newCard];
     setCards(updatedCards);
```
sabellius: Why cards are used as a local state? It's an array under the list object, so I not really sure whats going on here.

frontend/src/store/actions/board-actions.js
```diff
@@ -183,12 +183,21 @@ export async function moveAllCards(
...
+    store.dispatch({
+      type: ADD_CARD.SUCCESS,
+      payload: { listId, addedCard, addCardToEnd },
+    });
   } catch (error) {
-    store.dispatch(setError("addCard", `Error adding card: ${error.message}`));
+    store.dispatch({ type: ADD_CARD.FAILURE, payload: error.message });
     throw error;
   }
 }
```
sabellius: This can use the `createAsyncAction` utility function.

frontend/package.json
```diff
@@ -10,7 +10,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@atlaskit/tokens": "^8.0.0",
+    "@atlaskit/tokens": "^8.4.0",
```
sabellius: Why did you update this?

frontend/src/components/List.jsx
LiorRey: There was a UI bug in which the scroll did not go all the way up/down in order to show the new card that was just added at the top/bottom.
This nested call to `requestAnimationFrame` solves that, and is more concise than using `useEffect` or `setTimeout` for that matter.
As I understood, the first `requestAnimationFrame` is to wait for React to update state,
and the second `requestAnimationFrame` is to wait for DOM to render.

frontend/src/assets/styles/components/AddCardForm.css
snoxlax: I agree. I’ll start using them, and in some of my PRs where I touched the CSS, I tried to use the theme and its colors.

frontend/src/components/AddCardForm.jsx
```diff
@@ -0,0 +1,66 @@
...
+  onCardAdded,
+  onHideAddCardForm,
+}) {
+  const [title, setTitle] = useState("");
+  const boardId = useSelector(state => state.boards.board._id);
+  const inputRef = useRef();
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, []);
```
snoxlax, line 21: Where is this used? do we need this ref?

frontend/src/components/List.jsx
```diff
@@ -54,40 +54,42 @@ export function List({
...
+  function handleHideAddCardForm() {
     setActiveAddCardListId(null);
   }
 
-  function handleAddCard() {
-    if (!newCardTitle) {
-      handleHideAddCard();
-      return;
-    }
+  function handleCardAdded(addedCard) {
```
snoxlax: It was handleAddCard why is it now handleCardAdded??

frontend/src/components/List.jsx
snoxlax: Yes, this only updates the local state and doesn’t use the actual addCard action. why?

frontend/src/components/List.jsx
```diff
@@ -54,40 +54,42 @@ export function List({
     handleClose();
   }
 
-  function handleShowAddCard() {
-    setNewCardTitle("");
+  function handleDeleteList() {
+    onRemoveList(list.id);
+    handleClose();
+  }
```
snoxlax: I don’t understand this one..
So `handleDeleteList()` is prop-drilled to `ListActionsMenu`.
In `ListActionsMenu` the prop isn’t received..
And `onRemoveList` itself is coming from `BoardDetails`.
And in `BoardDetails` the function is empty.

And in general, why are we “half-implementing” delete list in a PR for `add-card-from-list-actions`?

frontend/package.json
LiorRey: I updated using the wrong command earlier.
Now changed back to `"@atlaskit/tokens": "^8.0.0"` after updating the `json` files directly from those on `main` branch. 👍

frontend/src/components/AddCardForm.jsx
LiorRey: Clicking to open **"List actions"** menu, triggers a **state change** (see the image below).
This makes the `AddCardForm` (that will appear at the top of the list, after clicking on "Add card") to have its inner text input lose focus - unlike the `AddCardForm` at the footer, where we can open it **without going through a state change**, and for which `autoFocus` was enough.

I had to use the above code, as `autoFocus` is not working for the upper `AddCardForm` because of that state change.

image

frontend/src/components/List.jsx
LiorRey: `handleAddCard` function was moved to the dedicated component `AddCardForm`, and it handles the call to the `addCard` action, and the logic to save the new card in the data.

At the end of `handleAddCard`, we have a callback to a the function `handleCardAdded` in `List`, that handles the appearance and render of the list, with the new card to be displayed in it.

Different functions with different purposes in different locations.

frontend/src/components/List.jsx
LiorRey: Probably I moved this function inside its file, or it was fetched from main, I did not dealt with deleting list at all.

frontend/src/components/List.jsx
LiorRey: Usage of old code. Removed and fixed 👍

frontend/src/store/actions/board-actions.js
LiorRey: Done 👍

frontend/src/assets/styles/components/AddCardForm.css
LiorRey: Fixed and done 👍

frontend/src/components/List.jsx
snoxlax: You can check the main branch it wasnt from main it was added form this pr

frontend/src/services/board/board-service-local.js
```diff
@@ -578,27 +596,6 @@ async function archiveAllCardsInList(boardId, listId) {
...
-      }
-    );
-
-    return updatedCardLabels;
-  } catch (error) {
-    console.error("Cannot update card labels:", error);
-    throw error;
-  }
-}
-
```
snoxlax, line 601: Why?
you just moved the function to a diffrent location why was it necessary?

frontend/src/services/board/index.js
```diff
@@ -16,17 +16,6 @@ function getEmptyBoard() {
...
-  return {
-    id: crypto.randomUUID(),
-    title: "",
-    description: "",
-    labels: [],
-    createdAt: null,
-    archivedAt: null,
-  };
-}
-
```
snoxlax, line 29: I dont understand this change 
At the beginning, I thought we wanted to keep this file because it had `getEmptyCard` and `getEmptyList`, but now it only has `getEmptyBoard` slowly this file is becoming empty
so what was the purpose of it?
@LiorRey @sabellius

frontend/src/store/actions/board-actions.js
```diff
@@ -183,15 +183,7 @@ export async function moveAllCards(
...
-export async function addCard(boardId, card, listId) {
-  try {
-    const newCard = await boardService.addCard(boardId, card, listId);
-    store.dispatch(addCardAction(newCard, listId));
-  } catch (error) {
-    store.dispatch(setError("addCard", `Error adding card: ${error.message}`));
-    throw error;
-  }
-}
+export const addCard = createAsyncAction(ADD_CARD, boardService.addCard, store);
```
snoxlax, line 186: Why the format is diffrent on this one?

frontend/src/services/board/board-service-local.js
```diff
@@ -169,6 +186,7 @@ export async function editCard(boardId, card, listId) {
 export async function deleteCard(boardId, cardId, listId) {
   try {
     const board = await getById(boardId);
+    if (!board) throw new Error("Board not found");
```
snoxlax: Why this was added?

frontend/src/services/board/board-service-local.js
LiorRey: Merge and conflicts resulted from pulling from main caused this.
Moved it up and now it's back at its place 👍

frontend/src/store/actions/board-actions.js
LiorRey: After this review was posted, we decided in our discussion that it can stay like this 👍

frontend/src/services/board/board-service-local.js
LiorRey: Removed 👍

frontend/src/services/board/index.js
LiorRey: After this review was posted, @snoxlax and I decided that it makes more sense to have `getEmptyCard` in the service, right about `addCard` function, instead of in `index.js`. as `getEmptyList` is also in the service 👍

frontend/src/components/List.jsx
LiorRey: Ghost-code that somehow remained there unused.
Removed `onRemoveList` and `handleDeleteList` from `BoardDetails` and `List`👍

frontend/src/components/List.jsx
```diff
@@ -54,40 +48,31 @@ export function List({
...
+        scrollToBottom,
+      })
     );
   }
 
-  function handleHideAddCard() {
-    setNewCardTitle("");
+  function handleHideAddCardForm() {
     setActiveAddCardListId(null);
   }
```
snoxlax: What dose this function do? its unclear by the naming

frontend/src/components/List.jsx
```diff
@@ -108,9 +93,19 @@ export function List({
...
+      <div className="list-add-card-header">
+        {isAddingCard && !addCardToEnd && (
+          <AddCardForm
+            listId={list.id}
+            addCardToEnd={addCardToEnd}
+            onCardAdded={scrollToAddedCard}
+            onHideAddCardForm={handleHideAddCardForm}
+          />
+        )}
+      </div>
```
snoxlax: Here the div of `list-add-card-header` will always remain as a dom element if isAddingCard is false it will still be there and empty

frontend/src/services/board/board-service-local.js
snoxlax: Its still here

frontend/src/store/actions/board-actions.js
snoxlax: No, we talked about it and it shouldnt be like this
It should look like this 
```js
export const editCard = createAsyncAction(
  EDIT_CARD,
  boardService.editCard,
  store
);
```
for example
the current format doesnt match our Prettier rules and i dont know why exactly?

frontend/src/services/board/board-service-local.js
LiorRey: Ohh I looked at line 189 and removed the check from `addCard` that I worked on.
Now I removed from `deleteCard` as well 👍

frontend/src/store/actions/board-actions.js
LiorRey: I save and it stays like that. Probably the `addCard` line is somewhat shorter than that of `editCard` for Prettier to break it.

frontend/src/components/List.jsx
LiorRey: It is used when closing the AddCardForm component.
Because it consists only one line, I removed the function and now I use only `setActiveAddCardListId(null)` where `handleHideAddCardForm` was called before 👍

frontend/src/components/List.jsx
LiorRey: Fixed 👍

.vscode/settings.json
snoxlax, line 1: can we remove this file?

snoxlax (approved):
only remove the [‎.vscode/settings.json](https://github.com/LiorRey/ca-final-project/pull/59/files#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357)

other than this Approved

.vscode/settings.json
LiorRey: Done 👍

sabellius (approved):

PR #74, opened 2025-12-05
bugfix/fix-missing-attributes-un-full-board-details

sabellius:
This pr fixes an issue where some properties that contain subdocuments array, and for some reason are empty/undefined, they don't appear in the response of the `getFullBoardDetails` action. This is caused by using lean, which removes virtual properties. There is also possible issue with this being nested virtual properties: boards -> lists -> cards.


snoxlax (approved):
Approved looks nice!

PR #82, opened 2025-12-09
Feature/catc 140 implement board index component

LiorRey:
## 🎯 Description

This PR improves the Board search experience by making the Header search input context-aware and navigation-friendly.
The search term now behaves consistently across routes and browser navigation, ensuring users see the same filtered boards when navigating back from a board to the boards list.

## 🔧 Changes

* 🔍 Context-aware Header search
The Header search input now adapts its behavior based on the current route (BoardIndex vs BoardDetails).

* 🔄 Preserved search state on browser back navigation
When navigating from a board back to the boards list using the browser’s Back button, the previous search term is restored and applied.

* 🧭 Improved navigation consistency
Prevents mismatches where filtered boards are shown but the search input appears empty (or vice versa).

* 🧼 Reduced unnecessary state resets
Removed aggressive clearing of the board search state on route changes, allowing React Router history to behave naturally.

## 📝 Notes

* Search state is intentionally preserved when navigating between BoardIndex and BoardDetails to match user expectations.

* The Header component determines behavior based on the current route instead of relying on additional Redux reducers.

* This lays groundwork for future route-specific Header behavior without increasing global state complexity.


frontend/src/assets/styles/setup/var.css
```diff
@@ -22,6 +22,8 @@
   --t-red: rgba(255, 0, 0, 0.5);
   --t-green: rgba(0, 255, 0, 0.5);
   --t-purple: rgba(128, 0, 128, 0.5);
+  --t-dark-gray: rgba(255, 255, 255, 0.16);
+  --dynamic-text-transparent: var(--t-dark-gray);
```
snoxlax: This is using the same color value, just referenced through a different variable name.

LiorRey: I wanted to use the `--dynamic-text-transparent` variable that I did not found on our system, but was on real Trello.
It just wraps the `--t-dark-gray` var, but I would like to have both dark gray transparent color always available, and also use the "--dynamic-text-transparent" when it appears.

snoxlax: Is `--t-dark-gray` actually used anywhere? I did not see it referenced in the code

snoxlax (approved):
Looks good!

frontend/src/assets/styles/setup/var.css
LiorRey: By itself no, but gray is a common color, and dark gray is used in real Trello.
Now only `--dynamic-text-transparent` uses it (which is used as well), but maybe we will use it directly in the future, so I think it should be kept.

LiorRey:
After getting approved in Discord - I shall merge! 🦾
image

PR #85, opened 2025-12-13
Attempted fix for rendering issue on BoardDetails and a simple BoardIndex

snoxlax:
## 📝 Notes

* This PR addresses the immediate rendering and ESLint issues; however, the proposed fix is not viable long-term.
* With this approach, lists do not update correctly when cards, labels, or other entities are edited.
* The application relies on a global store as the single source of truth. Any change must propagate from the store to the board to keep state consistent.
* As a result, lists will currently render twice. This is a known and accepted limitation for now, pending a proper architectural solution.

## 🖥️ Demo

<details>

  <p align="center">
  </p>

</details>

### DIFF
--- a/frontend/src/assets/styles/components/Board.css
+++ b/frontend/src/assets/styles/components/Board.css
@@ -1,5 +1,67 @@
 .board-index {
   height: 100%;
+  padding: 2rem;
+  overflow-y: auto;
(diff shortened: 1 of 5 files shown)


snoxlax:
It does not work as well as I initially expected, so it is here for reference only.

PR #86, opened 2025-12-14
feature/CATC_144-connect-frontend-with-backend

sabellius:
## 🎯 Description
This PR connects the Trello clone frontend with the backend server, replacing local storage with REST API calls for real data persistence and multi-user functionality.

## 🔧 Changes
- Connected both auth and board services (remote versions) with the backend.

## 📝 Notes
- Made some fixes to bugs i found in the backend/frontend along the way.
- Some function are still commented out and wont work in at all as they aren't implemented in the server yet.
- Many of the implementations are breaking changes  as some function signatures changed and using the local service will not be possible anymore.
- Some state changes in the frontend were made to work with different responses from the server from what we worked with locally. for example when we move a list, we don't want to query all the board lists all over, and instead we only return the moved list, with which hte frontend will know whether to filter it out or to resort the lists array (no need to send data the frontend already has, and this also is consistent with REST way of doing things).
- The following functoin currently wont work, as most of them are not yet implemented in the backend. Will work on them ASAP.

### DIFF
--- a/frontend/src/store/reducers/board-reducer.js
+++ b/frontend/src/store/reducers/board-reducer.js
@@ -144,11 +144,11 @@ const handlers = {
     board: {
       ...state.board,
       lists: state.board.lists.map(list =>
-        list._id !== action.payload.card.listId
+        list._id !== action.payload.listId
           ? list
           : {
               ...list,
-              cards: sortByPosition([...list.cards, action.payload.card]),
+              cards: sortByPosition([...list.cards, action.payload]),
             }
       ),
     },
--- a/frontend/src/store/reducers/board-reducer.js
+++ b/frontend/src/store/reducers/board-reducer.js
@@ -200,7 +200,17 @@ const handlers = {
     loading: { ...state.loading, [MOVE_CARD.KEY]: false },
     board: {
       ...state.board,
-      lists: action.payload,
+      lists: state.board.lists.map(list =>
+        list._id === action.payload.listId
+          ? {
+              ...list,
+              cards: sortByPosition([...list.cards, action.payload]),
+            }
+          : {
+              ...list,
+              cards: list.cards.filter(card => card._id !== action.payload._id),
+            }
+      ),
     },
   }),
   ...createAsyncHandlers(CREATE_LABEL, CREATE_LABEL.KEY),
--- a/frontend/src/store/reducers/board-reducer.js
+++ b/frontend/src/store/reducers/board-reducer.js
@@ -160,12 +160,12 @@ const handlers = {
     board: {
       ...state.board,
       lists: state.board.lists.map(list =>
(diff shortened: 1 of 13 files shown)


frontend/src/store/reducers/board-reducer.js
```diff
@@ -109,11 +109,22 @@ const handlers = {
...
+    if (state.board._id !== action.payload.boardId) {
+      newLists = state.board.lists.filter(
+        list => list._id !== action.payload._id
+      );
+    } else {
+      newLists = state.board.lists.map(list =>
+        list._id === action.payload._id ? { ...list, ...action.payload } : list
+      );
+      newLists = sortByPosition(newLists);
+    }
```
snoxlax, line 122: can we put this inside the service instead of here?

frontend/src/store/reducers/board-reducer.js
```diff
@@ -46,7 +46,7 @@ const handlers = {
     loading: { ...state.loading, [ADD_LIST.KEY]: false },
     board: {
       ...state.board,
-      lists: [...state.board.lists, action.payload],
+      lists: sortByPosition([...state.board.lists, action.payload]),
```
snoxlax, line 49: Is it possible to move this into the service? I think that would be cleaner

frontend/src/store/reducers/board-reducer.js
```diff
@@ -133,7 +144,12 @@ const handlers = {
     board: {
       ...state.board,
       lists: state.board.lists.map(list =>
-        list._id === action.payload._id ? action.payload : list
+        list._id !== action.payload.listId
+          ? list
+          : {
+              ...list,
+              cards: sortByPosition([...list.cards, action.payload]),
```
snoxlax, line 160: and this

frontend/src/store/reducers/board-reducer.js
sabellius: We could, but I think the purpose of the service is to abstract the interaction with the back-end, do we really want to mix the UI state manipulation in there as well?

frontend/src/store/reducers/board-reducer.js
sabellius: Again, this probably should go into the service because it's not related to the logic of the service, the services' purpose is to fetch or send data to the server and receive whatever the server responds with. For example in this case, we add a list, the service send some data to the server, and receives a newly created list object. The service doesn't and shouldn't know about other things and contexts such as lists how how the returned data should be manipulated in the UI. If the reducer cases seem too long to you, perhaps a better solution would be to extract some of this logic into store utility/helper functions, but in terms of boundaries, I don't think we should mix this we the board service. What do you think?

snoxlax (approved):
Looks good!

PR #87, opened 2025-12-14
feature/CATC_143-implement-copy-list-in-the-backend

sabellius:
## 🎯 Description
Implements backend Copy List functionality, allowing users to duplicate lists with options to copy cards, specify target board and position.

## 🔧 Changes
- 🚀 Added `copyList` service with options: `copyCards`, `targetBoardId`, `targetIndex`, `title`
- 🎯 Added `copyList` controller for HTTP request handling
- 🛣️ Added `POST /:id/copy` route with auth/authorization
- 💾 Returns populated list with cards for frontend state management

## 📝 Notes
- Creates deep copies of cards when `copyCards: true`
- By default the copied list is copied right next to the original list, just like in the real Trello.

### DIFF
--- a/backend/src/controllers/list-controller.js
+++ b/backend/src/controllers/list-controller.js
@@ -49,3 +49,10 @@ export async function deleteList(req, res) {
 
   res.status(200).json({ id: deletedList._id });
 }
+
+export async function copyList(req, res) {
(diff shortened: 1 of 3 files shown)


snoxlax (approved):
Looks good!

PR #90, opened 2025-12-14
Added assignee control on cards with remote service

snoxlax:
## 🎯 Description

This PR introduces full assignee management for cards using the remote service. Users can now add and remove assignees directly from a card, with multiple entry points in the UI.

## 🔧 Changes

* Added support for assigning members to a card via the remote service
* Enabled removal of assignees from a card
* Extended the card modal to allow managing assignees in two ways:
  * From the **Add Member** section inside the modal
  * From the **“+” members icon**, using the assignee selection menu
* Implemented avatar rendering using member initials with a unique, consistent color per member

## 📝 Notes

* Avatars currently display user initials and a generated color.
* Once attachments and image uploads are supported, avatars will be updated to use profile images when available.
* If no image is provided, the system will gracefully fall back to the current initials-and-color avatar implementation.

---

## 🖥️ Demo

<details>

  <p align="center">
  </p>

</details>

### DIFF
--- a/frontend/src/components/AddMemberMenu.jsx
+++ b/frontend/src/components/AddMemberMenu.jsx
@@ -24,12 +24,16 @@ export function AddMemberMenu({
         !cardAssignees.some(assignee => assignee.userId === member.userId)
     );
 
-    const filteredMembers = boardMembersList.filter(member =>
-      member.fullname?.toLowerCase().includes(searchTerm.toLowerCase())
+    const filteredMembers = boardMembersList.filter(
+      member =>
+        member.fullname?.toLowerCase().includes(searchTerm.toLowerCase()) ||
+        member.username?.toLowerCase().includes(searchTerm.toLowerCase())
     );
 
-    const filteredAssignees = cardAssignees.filter(assignee =>
-      assignee.fullname?.toLowerCase().includes(searchTerm.toLowerCase())
+    const filteredAssignees = cardAssignees.filter(
+      assignee =>
+        assignee.fullname?.toLowerCase().includes(searchTerm.toLowerCase()) ||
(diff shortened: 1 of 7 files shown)


frontend/src/components/AddMemberMenu.jsx
```diff
@@ -1,84 +1,54 @@
...
-    const cardMembersList = filtered.filter(member =>
-      cardMemberIds.includes(member._id)
+    const filteredMembers = boardMembersList.filter(member =>
+      member.fullname?.toLowerCase().includes(searchTerm.toLowerCase())
     );
-    const boardMembersList = filtered.filter(
-      member => !cardMemberIds.includes(member._id)
+
+    const filteredAssignees = cardAssignees.filter(assignee =>
+      assignee.fullname?.toLowerCase().includes(searchTerm.toLowerCase())
```
sabellius: Does real Trello support doing this with username as well?

frontend/src/components/CardModal.jsx
```diff
@@ -133,14 +121,19 @@ export function CardModal({
...
-                    {cardMembers && cardMembers.length > 0 && (
+                    {card.assignees && card.assignees.length > 0 && (
                       <>
-                        {cardMembers.map(member => (
-                          <Avatar key={member._id} user={member} size={36} />
+                        {card.assignees.map(assignee => (
+                          <Avatar
+                            key={assignee.userId}
+                            user={assignee}
+                            size={36}
```
sabellius, line 135: Should it really be this big? Even if it works like this perhaps it's better to lower this so the avatar with the number will be displayed even with a small dataset?

sabellius (approved):
LGTM :)

PR #91, opened 2025-12-15
Added loggedIn user icon and menu

snoxlax:
## 🎯 Description

Added a user menu that appears when clicking the logged-in user's avatar in the header. The menu displays user information (avatar, name, and email) and provides a logout option.

## 🔧 Changes

- **User Menu Component**: Created a new `UserMenu` component that displays user account information and menu actions
- **User Menu Styling**: Added dedicated `UserMenu.css` file with styles for the menu, user info section, and menu items
- **Clickable Avatar**: Made the user avatar in the header clickable to open the user menu
- **Menu Structure**: Implemented menu items using the same pattern as `BoardMenu` for consistency
- **Logout Functionality**: Added logout menu item with icon that navigates to home page after logout
- **Component Separation**: Extracted user menu logic from `Header` component into a separate reusable component

## 📝 Notes

- The menu uses the existing `PopoverMenu` component for consistent UI
- Menu items follow the same structure as `BoardMenu` for maintainability
- User email is displayed with a darker tone for better visual hierarchy
- The menu automatically closes after logout action
- All styles are organized in a dedicated CSS file separate from header styles

## 🖥️ Demo

<details>

  <p align="center">
  </p>

</details>

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -1,3 +1,4 @@
+import { useEffect } from "react";
 import { Routes, Route, useLocation } from "react-router-dom";
 import { Header } from "./components/Header";
 import { UserMessage } from "./components/UserMessage";
@@ -12,10 +13,15 @@ import { LoginPage } from "./pages/LoginPage";
 import { SignupPage } from "./pages/SignupPage";
 import { ThemeComparison } from "./pages/ThemeComparison";
 import { UserDetails } from "./pages/UserDetails";
+import { restoreSession } from "./store/actions/auth-actions";
 
 export function App() {
   const location = useLocation();
   const backgroundLocation = location.state?.backgroundLocation;
+
+  useEffect(() => {
+    restoreSession();
+  }, []);
 
   return (
     <div className="main-container">
--- a/frontend/src/components/UserMenu.jsx
+++ b/frontend/src/components/UserMenu.jsx
@@ -1,6 +1,6 @@
 import { useMemo } from "react";
 import { useNavigate } from "react-router";
-import { logout } from "../store/actions/user-actions";
+
 import { Avatar } from "./ui/Avatar";
 import { PopoverMenu } from "./ui/PopoverMenu";
 import ListItem from "@mui/material/ListItem";
@@ -9,6 +9,7 @@ import ListItemIcon from "@mui/material/ListItemIcon";
 import ListItemText from "@mui/material/ListItemText";
 import MuiList from "@mui/material/List";
 import { ExitToApp } from "@mui/icons-material";
+import { logout } from "../store/actions/auth-actions";
 
 function UserMenuItem({ item, onItemClick }) {
   return (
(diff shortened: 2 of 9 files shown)


sabellius (approved):

frontend/src/components/UserMenu.jsx
```diff
@@ -0,0 +1,87 @@
+import { useMemo } from "react";
+import { useNavigate } from "react-router";
+import { logout } from "../store/actions/user-actions";
```
sabellius: I looked at this again, and the function used here is from the starter code. We discussed this before, and agreed that auth related logic should reside in the auth service, So I think it would be better moving this logic to the auth service. What do you think?

P.S. - please make sure the the correct action is used (and perhaps use the store helpers where possible as well)

sabellius (commented):
Please see my comment.

sabellius (changes requested):
For some reason it keeps it approved.

frontend/src/components/UserMenu.jsx
snoxlax: we have our own logout? i just didnt knew what to use here

sabellius: We currently don't have an implementation of logout `logout` in the `auth-actions.js`, but it can easiy be done using the `auth-service-remote.js` function along with the redux patterns we used throughout the app.

frontend/src/store/actions/auth-actions.js
```diff
@@ -34,3 +40,35 @@ export async function login(credentials) {
...
+    await userService.logout();
+    store.dispatch({ type: LOGOUT.SUCCESS });
+  } catch (error) {
+    store.dispatch({
+      type: LOGOUT.FAILURE,
+      payload: error.message,
+      key: LOGOUT.KEY,
+    });
+    throw error;
+  }
```
sabellius: You can use the helper if there is nothing different in the async action.

frontend/src/App.jsx
```diff
@@ -12,11 +13,16 @@ import { LoginPage } from "./pages/LoginPage";
...
+import { restoreSession } from "./store/actions/auth-actions";
 
 export function App() {
   const location = useLocation();
   const backgroundLocation = location.state?.backgroundLocation;
 
+  useEffect(() => {
+    restoreSession();
+  }, []);
+
```
sabellius, line 25: I'm currently working on this part, this alone doesn't work for several reasons:

1. This is performed asynchronously, meaning that the route (even if protected), is loaded before this session check is finished - resulting in loading a page that the user is not supposed to access.

2. The is no route protection, so in case a user will enter a board page directly, the page will be rendered, but the user might not be logged in and you won't get a chance to redirect him to a public page such as logic, or homepage.

I'm currently working on this exact part, so let me know if you want to handle this properly or if you wish to merge this PR without this and let me continue working on this.

sabellius (approved):

PR #95, opened 2025-12-16
Added cover color control for cards

snoxlax:
## 🎯 Description

This PR initializes Cloudinary integration to support future attachment management and introduces cover controls for cards, enabling visual customization at the card level.

## 🔧 Changes

* **Cloudinary Initialization**
  Set up Cloudinary for future attachment and media handling.

* **Card Cover Controls**
  Added support for card covers using colors, configurable directly from the card modal.

* **Cover Style Customization**
  Enabled changing the visual style of the card cover.

* **Board-wide Sync**
  Cover changes are immediately reflected across the entire board for all cards.

## 📝 Notes

* Cloudinary is currently initialized for infrastructure and future use; attachment uploads are not yet exposed to the UI.
* The cover feature supports color-based covers at this stage.

## 🖥️ Demo

<details>

  <p align="center">
  </p>

</details>

### DIFF
--- a/backend/src/controllers/card-controller.js
+++ b/backend/src/controllers/card-controller.js
@@ -27,6 +27,15 @@ export async function updateCard(req, res) {
   res.json({ card });
 }
 
+export async function upsertCardCover(req, res) {
+  const { id } = req.params;
+  const { color, img, textOverlay } = req.body;
+  const card = await cardService.upsertCardCover(id, color, img, textOverlay);
+  if (!card) throw createError(404, "Card not found");
+
+  res.status(200).json({ card });
+}
+
 export async function deleteCard(req, res) {
   const card = await cardService.deleteCard(req.params.id);
   if (!card) throw createError(404, "Card not found");
--- a/backend/src/controllers/card-controller.js
+++ b/backend/src/controllers/card-controller.js
@@ -27,10 +27,9 @@ export async function updateCard(req, res) {
   res.json({ card });
 }
 
-export async function upsertCardCover(req, res) {
+export async function updateCover(req, res) {
   const { id } = req.params;
-  const { color, img, textOverlay } = req.body;
-  const card = await cardService.upsertCardCover(id, color, img, textOverlay);
+  const card = await cardService.updateCover(id, req.body);
   if (!card) throw createError(404, "Card not found");
 
   res.status(200).json({ card });
--- a/backend/src/models/Card.js
+++ b/backend/src/models/Card.js
@@ -54,6 +54,20 @@ const cardSchema = new mongoose.Schema(
       required: true,
       trim: true,
     },
+    cover: {
+      img: {
+        type: String,
+        default: null,
+      },
+      color: {
+        type: String,
+        default: null,
+      },
+      textOverlay: {
+        type: Boolean,
+        default: false,
+      },
+    },
     description: {
       type: String,
       default: "",
--- a/backend/src/services/card-service.js
+++ b/backend/src/services/card-service.js
@@ -200,6 +200,26 @@ export async function removeCardAssignee(cardId, userId) {
   );
 }
 
+export async function upsertCardCover(cardId, color, img, textOverlay) {
+  const card = await Card.findByIdAndUpdate(
+    cardId,
+    {
+      $set: {
+        "cover.img": img !== undefined ? img : null,
+        "cover.color": color !== undefined ? color : null,
+        "cover.textOverlay": textOverlay !== undefined ? textOverlay : false,
+      },
+    },
+    {
+      new: true,
+      runValidators: true,
+    }
+  );
+
+  if (!card) throw createError(404, "Card not found");
+  return card;
+}
+
 export async function copyCard(cardId, copyOptions = {}) {
   const {
     copyLabels = false,
--- a/backend/src/services/card-service.js
+++ b/backend/src/services/card-service.js
@@ -200,14 +200,15 @@ export async function removeCardAssignee(cardId, userId) {
   );
 }
 
-export async function upsertCardCover(cardId, color, img, textOverlay) {
+export async function updateCover(cardId, coverData) {
+  const { color, img, textOverlay } = coverData;
   const card = await Card.findByIdAndUpdate(
     cardId,
     {
       $set: {
-        "cover.img": img !== undefined ? img : null,
-        "cover.color": color !== undefined ? color : null,
-        "cover.textOverlay": textOverlay !== undefined ? textOverlay : false,
+        "cover.img": img,
+        "cover.color": color,
+        "cover.textOverlay": textOverlay,
       },
     },
     {
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,6 +11,8 @@
   },
   "dependencies": {
     "@atlaskit/tokens": "^8.0.0",
+    "@cloudinary/react": "^1.14.3",
+    "@cloudinary/url-gen": "^1.22.0",
     "@emotion/react": "^11.14.0",
     "@emotion/styled": "^11.14.1",
     "@hello-pangea/dnd": "^18.0.1",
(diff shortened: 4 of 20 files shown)


backend/src/controllers/card-controller.js
```diff
@@ -27,6 +27,15 @@ export async function updateCard(req, res) {
   res.json({ card });
 }
 
+export async function upsertCardCover(req, res) {
+  const { id } = req.params;
+  const { color, img, textOverlay } = req.body;
+  const card = await cardService.upsertCardCover(id, color, img, textOverlay);
```
sabellius: IMHO, I think it would be better to pass an object to the service function, because if you'd like to add additional fields to the cover object, you will only need to update the passed object instead the function signature and the function invocations.

backend/src/services/card-service.js
```diff
@@ -200,6 +200,26 @@ export async function removeCardAssignee(cardId, userId) {
   );
 }
 
+export async function upsertCardCover(cardId, color, img, textOverlay) {
```
sabellius: Why did you choose to use the term upsert? Is the purpose of this function is to create a card if it doesn't find one with the `cardId` specified in the URL? If yes, how does `findOneAndUpdate` handles this? If no, please clarify.

backend/src/services/card-service.js
```diff
@@ -200,6 +200,26 @@ export async function removeCardAssignee(cardId, userId) {
...
+  const card = await Card.findByIdAndUpdate(
+    cardId,
+    {
+      $set: {
+        "cover.img": img !== undefined ? img : null,
+        "cover.color": color !== undefined ? color : null,
+        "cover.textOverlay": textOverlay !== undefined ? textOverlay : false,
+      },
```
sabellius, line 212: I believe this will overwrite any existing values in those fields, for example if there was a `cover.color` before this operation, and in the request no color was provided, the existing color will be overwritten with `null`. Is this the intended behavior?

Also, I see that you perform the same existence check for each of the attributes, what if you'll want to add additional attributes in the future? Will they need the same check? If so, Would you agree you could use some helper function here or in the controller that filters out undefined attributes so it would scale a litter better and be more DRY?

frontend/src/components/AttachmentMenu.jsx
```diff
@@ -0,0 +1,134 @@
+import { useState } from "react";
+import { PopoverMenu } from "./ui/PopoverMenu";
+import "../assets/styles/components/AttachmentMenu.css";
```
sabellius: Isn't this supposed to be imported in `main.css`?

frontend/src/components/Card.jsx
```diff
@@ -76,96 +79,122 @@ export function Card({
...
-                    </div>
-                  ))}
-                </div>
+            <Box
+              className="card-content-container"
+              onClick={handleClickCard}
+              sx={
+                card.cover?.textOverlay
+                  ? { backgroundColor: card.cover.color }
+                  : undefined
```
sabellius: I think a short-cercuit is more intuitive here - `sx={card.cover?.textOverlay && { backgroundColor: card.cover.color }}` especially since you don't really need an "else" here. What do you think?

frontend/src/components/AttachmentMenu.jsx
```diff
@@ -0,0 +1,134 @@
...
+import { upsertCardCover } from "../store/actions/board-actions";
+import { COVER_COLORS } from "../services/board/board-backgrounds";
+
+export function AttachmentMenu({
+  card,
+  anchorEl,
+  isAttachFileMenuOpen,
+  onCloseAttachFileMenu,
+}) {
+  const [textOverlay, setTextOverlay] = useState(card.cover?.textOverlay);
```
sabellius: Should this value be a `Boolean`? current implementation allows for `undefined` as well. Can this break anything?

frontend/src/components/AttachmentMenu.jsx
```diff
@@ -0,0 +1,134 @@
...
+  function handleCloseMenu() {
+    onCloseAttachFileMenu();
+  }
+
+  function handleTextOverlay(overlay) {
+    setTextOverlay(overlay);
+    upsertCardCover(card._id, {
+      color: selectedColor?.value,
+      textOverlay: overlay,
+    });
```
sabellius: If this updates the local state, but also updates the card object which is passed as props to this component, doesn't this cause multiple renders? Shouldn't firing the redux action be enough to update the local state via prop change enough? 

The same question goes for `handleColorSelect` and `handleRemoveCover`.

frontend/src/components/AttachmentMenu.jsx
```diff
@@ -0,0 +1,134 @@
...
+        vertical: "bottom",
+        horizontal: "left",
+      }}
+      transformOrigin={{
+        vertical: "top",
+        horizontal: "left",
+      }}
+      paperProps={{ sx: { mt: 1 } }}
+      sx={{
+        zIndex: theme => theme.zIndex.modal + 1,
```
sabellius: I've noticed that the `CardPopover` component has the same `z-index` value, can this cause issues in the ui?

frontend/src/services/board/board-service-remote.js
```diff
@@ -83,6 +84,15 @@ async function editCard(_boardId, card, _listId) {
   return data.card;
 }
 
+async function upsertCardCover(cardId, coverData) {
+  const data = await httpService.put(`cards/${cardId}/cover`, {
+    img: coverData.img,
+    color: coverData.color,
+    textOverlay: coverData.textOverlay,
```
sabellius: Minor thing, but if the `coverData` is exactly the same object as the one you're building here, you can just pass it instead.

backend/src/models/Card.js
```diff
@@ -54,6 +54,20 @@ const cardSchema = new mongoose.Schema(
...
+      },
+      color: {
+        type: String,
+        default: null,
+      },
+      textOverlay: {
+        type: Boolean,
+        default: false,
+      },
+    },
```
sabellius, line 70: It's not a must, but I think generating a migration to populate some cards with covers would be nice, both for development and for demo data.

frontend/package.json
```diff
@@ -11,6 +11,8 @@
   },
   "dependencies": {
     "@atlaskit/tokens": "^8.0.0",
+    "@cloudinary/react": "^1.14.3",
+    "@cloudinary/url-gen": "^1.22.0",
```
sabellius, line 15: I can see the dependencies, but can't find where they are used. is this a setup for uploading files? If so, Is there any advancement on this subject?

frontend/src/components/AttachmentMenu.jsx
```diff
@@ -0,0 +1,134 @@
...
+        )}
+
+        <div className="cover-section">
+          <label className="cover-section-label">Colors</label>
+          <div className="cover-colors-grid">
+            {COVER_COLORS.map(color => (
+              <button
+                key={color.name}
+                className={`cover-color-swatch ${
+                  selectedColor === color.name ? "selected" : ""
```
sabellius: Does this work? `selectedColor` seems to be an object in the `useState` and here you compare it with a string.

sabellius (changes requested):
Please see my comments.

backend/src/services/card-service.js
snoxlax: i wanted to make this for cover not for card if there is no cover add it or update if it exists

frontend/package.json
snoxlax: Yes this cover is the start and i will continue in another pr beacuse this one is already to big

frontend/src/components/AttachmentMenu.jsx
snoxlax: No it didn't but now i see why. Thanks, will fix

backend/src/services/card-service.js
sabellius: I think the terminology is incorrect, usually upsert is used for object/documents, not for specific properties of a document. a card can either have a cover or not - so a better term for this in my opinion would be something like `setCardCover`, or `updateCardCover`. Your thoughts? Please let me know if I'm missing something here.

backend/src/models/Card.js
sabellius: @snoxlax ?

sabellius (approved):

PR #96, opened 2025-12-17
Add attachment-based card covers

snoxlax:
## 🎯 Description

This PR introduces **attachment-based card covers**, enabling cards to use uploaded attachments as cover images.

All uploads are handled through a **secure, backend-signed Cloudinary flow**, ensuring attachments are validated and authorized before being stored and applied as card covers.

## 🔧 Changes

* Add Cloudinary integration and environment variable initialization on the backend
* Introduce a dedicated backend Cloudinary service and attachment upload/signing service
* Register Cloudinary-related routes in the backend index
* Fix and standardize function naming across attachment and remote services
* Add attachment menu options and related UI styling
* Enable card cover updates (color and image) via the card modal
* Apply styling updates for card cover color and image rendering

## 📝 Notes

* Uploads use **signed Cloudinary requests** generated by the backend
* No Cloudinary secrets are exposed to the frontend
* This PR focuses on the core attachment-to-cover flow; further refinements may follow

## 🖥️ Demo

<details>

  <p align="center">
  </p>

</details>

### DIFF
--- a/backend/src/controllers/cloudinary-controller.js
+++ b/backend/src/controllers/cloudinary-controller.js
@@ -1,0 +1,16 @@
+import createError from "http-errors";
+import * as cloudinaryService from "../services/cloudinary-service.js";
+
+const ALLOWED_FOLDERS = new Set(["card-covers"]);
+
+export async function getUploadSignature(req, res) {
+  const folder = req.body?.folder;
+  if (!folder) throw createError(400, "folder is required");
+  if (!ALLOWED_FOLDERS.has(folder)) throw createError(400, "Invalid folder");
+
+  const signData = cloudinaryService.generateUploadSignature(folder);
+  if (!signData)
+    throw createError(500, "Failed to generate Cloudinary signature");
+
+  res.json({ signData });
+}
--- a/backend/src/controllers/upload-controller.js
+++ b/backend/src/controllers/upload-controller.js
@@ -1,5 +1,5 @@
 import createError from "http-errors";
-import * as cloudinaryService from "../services/cloudinary-service.js";
+import * as uploadService from "../services/upload-service.js";
 
 const ALLOWED_FOLDERS = new Set(["card-covers"]);
 
@@ -8,9 +8,8 @@ export async function getUploadSignature(req, res) {
   if (!folder) throw createError(400, "folder is required");
   if (!ALLOWED_FOLDERS.has(folder)) throw createError(400, "Invalid folder");
 
-  const signData = cloudinaryService.generateUploadSignature(folder);
-  if (!signData)
-    throw createError(500, "Failed to generate Cloudinary signature");
+  const signData = uploadService.generateUploadSignature(folder);
+  if (!signData) throw createError(500, "Failed to generate upload signature");
 
-  res.json({ signData });
+  res.status(200).json({ signData });
 }
--- a/backend/src/services/upload-service.js
+++ b/backend/src/services/upload-service.js
@@ -5,7 +5,7 @@ import { config } from "../config/env.js";
 function ensureConfigured() {
   const { cloudName, apiKey, apiSecret } = config.cloudinary || {};
   if (!cloudName || !apiKey || !apiSecret) {
-    throw createError(500, "Cloudinary is not configured");
+    throw createError(503, "Upload provider is not configured");
   }
 
   cloudinary.config({
--- a/frontend/src/components/AttachmentMenu.jsx
+++ b/frontend/src/components/AttachmentMenu.jsx
@@ -1,6 +1,8 @@
+import { useState } from "react";
 import { PopoverMenu } from "./ui/PopoverMenu";
-import { upsertCardCover } from "../store/actions/board-actions";
+import { updateCardCover } from "../store/actions/board-actions";
 import { COVER_COLORS } from "../services/board/board-backgrounds";
+import { attachmentService } from "../services/attachment-service";
 
 export function AttachmentMenu({
   card,
@@ -10,24 +12,52 @@ export function AttachmentMenu({
 }) {
   const textOverlay = card.cover?.textOverlay;
   const coverColor = card.cover?.color;
+  const coverImg = card.cover?.img;
+
+  const [isUploading, setIsUploading] = useState(false);
+  const [uploadError, setUploadError] = useState("");
 
   function handleCloseMenu() {
     onCloseAttachFileMenu();
   }
 
   function handleTextOverlay(overlay) {
-    upsertCardCover(card._id, {
+    updateCardCover(card._id, {
       color: coverColor,
+      img: coverImg,
       textOverlay: overlay,
     });
   }
 
   function handleColorSelect(color) {
-    upsertCardCover(card._id, { color: color.value, textOverlay });
+    updateCardCover(card._id, { color: color.value, textOverlay, img: null });
   }
 
   function handleRemoveCover() {
-    upsertCardCover(card._id, { color: null, textOverlay: false });
+    updateCardCover(card._id, { color: null, textOverlay: false, img: null });
+  }
+
+  async function handleUploadCover(ev) {
+    const file = ev.target.files?.[0];
+    if (!file) return;
+
+    try {
+      setIsUploading(true);
+      setUploadError("");
+
+      const uploaded = await attachmentService.uploadImage(file, "card-covers");
+
+      updateCardCover(card._id, {
+        img: uploaded.secure_url,
+        color: null,
+        textOverlay: textOverlay,
+      });
+    } catch (err) {
+      setUploadError(err?.message || "Upload failed");
+    } finally {
+      setIsUploading(false);
+      ev.target.value = "";
+    }
   }
 
   return (
@@ -46,7 +76,18 @@ export function AttachmentMenu({
       }}
     >
       <div className="attachment-menu-content">
-        {coverColor && (
+        <div className="cover-section">
+          <label className="cover-section-label">Upload</label>
+          <input type="file" accept="image/*" onChange={handleUploadCover} />
+          {isUploading && <div style={{ marginTop: 6 }}>Uploading...</div>}
+          {uploadError && (
+            <div style={{ marginTop: 6, color: "crimson", fontSize: 12 }}>
+              {uploadError}
+            </div>
+          )}
+        </div>
+
+        {(coverColor || coverImg) && (
           <div className="cover-section">
             <label className="cover-section-label">Style</label>
             <div className="cover-size-options">
@@ -58,7 +99,10 @@ export function AttachmentMenu({
                 className={`cover-size-preview ${
                   textOverlay === false ? "selected" : ""
                 }`}
-                style={{ backgroundColor: coverColor }}
+                style={{
+                  backgroundColor: coverImg ? undefined : coverColor,
+                  backgroundImage: coverImg ? `url(${coverImg})` : undefined,
+                }}
                 onClick={() => handleTextOverlay(false)}
               >
                 <CoverPreviewSkeleton className="cover-preview-container" />
@@ -71,7 +115,10 @@ export function AttachmentMenu({
                 className={`cover-size-preview ${
                   textOverlay === true ? "selected" : ""
                 }`}
-                style={{ backgroundColor: coverColor }}
+                style={{
+                  backgroundColor: coverImg ? undefined : coverColor,
+                  backgroundImage: coverImg ? `url(${coverImg})` : undefined,
+                }}
                 onClick={() => handleTextOverlay(true)}
               >
                 <CoverPreviewSkeleton className="cover-preview-container-transparent" />
@@ -80,7 +127,7 @@ export function AttachmentMenu({
           </div>
         )}
 
-        {coverColor && (
+        {(coverColor || coverImg) && (
           <button className="remove-cover-button" onClick={handleRemoveCover}>
             Remove cover
           </button>
--- a/frontend/src/components/AttachmentMenu.jsx
+++ b/frontend/src/components/AttachmentMenu.jsx
@@ -60,6 +60,12 @@ export function AttachmentMenu({
     }
   }
 
+  const coverStyle = coverImg
+    ? { backgroundImage: `url(${coverImg})` }
+    : coverColor
+    ? { backgroundColor: coverColor }
+    : undefined;
+
   return (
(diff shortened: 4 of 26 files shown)


frontend/src/components/AttachmentMenu.jsx
```diff
@@ -10,24 +12,52 @@ export function AttachmentMenu({
...
+      updateCardCover(card._id, {
+        img: uploaded.secure_url,
+        color: null,
+        textOverlay: textOverlay,
+      });
+    } catch (err) {
+      setUploadError(err?.message || "Upload failed");
+    } finally {
+      setIsUploading(false);
+      ev.target.value = "";
```
sabellius, line 59: The input is of `type="file"` so what does this do exactly?

backend/src/controllers/cloudinary-controller.js
```diff
@@ -0,0 +1,16 @@
...
+import * as cloudinaryService from "../services/cloudinary-service.js";
+
+const ALLOWED_FOLDERS = new Set(["card-covers"]);
+
+export async function getUploadSignature(req, res) {
+  const folder = req.body?.folder;
+  if (!folder) throw createError(400, "folder is required");
+  if (!ALLOWED_FOLDERS.has(folder)) throw createError(400, "Invalid folder");
+
+  const signData = cloudinaryService.generateUploadSignature(folder);
```
sabellius: I think it's a bad idea to name files and service with library/framework specific names. What if you would like to swap Cloudinary for a different upload service in the future? This is a great example where the [Dependency Inversion Pattern](https://en.wikipedia.org/wiki/Dependency_inversion_principle) can and probably should be used (similarily to how it's done in our front-end with the `http-service`). For example having something like a generic `uploadService` that can generate clients for both Cloudinary, AWS, or other providers.

backend/src/services/upload-service.js
```diff
@@ -0,0 +1,36 @@
...
+  const signature = cloudinary.utils.api_sign_request(
+    { timestamp, folder },
+    cloudinary.config().api_secret
+  );
+
+  return {
+    signature,
+    timestamp,
+    cloudName: cloudinary.config().cloud_name,
+    apiKey: cloudinary.config().api_key,
```
sabellius, line 33: Is this really something you should be sending in the response?

backend/src/controllers/cloudinary-controller.js
```diff
@@ -0,0 +1,16 @@
...
+  if (!signData)
+    throw createError(500, "Failed to generate Cloudinary signature");
```
sabellius: If a condition is wrapped over more than a single line, It's better to wrap it with curly braces, as this can lead to weird bugs.

Also, 500 is kind of vague, what are the possible causes that it could fail to generate a signature? From the service function signature, it seems there will always be some kind of object returned as the return value, so it seems that this condition will always evaluate to `false`.

backend/src/controllers/cloudinary-controller.js
```diff
@@ -0,0 +1,16 @@
...
+
+  res.json({ signData });
```
sabellius: status?

frontend/src/components/AttachmentMenu.jsx
```diff
@@ -10,24 +12,52 @@ export function AttachmentMenu({
...
+      setIsUploading(true);
+      setUploadError("");
+
+      const uploaded = await attachmentService.uploadImage(file, "card-covers");
+
+      updateCardCover(card._id, {
+        img: uploaded.secure_url,
+        color: null,
+        textOverlay: textOverlay,
+      });
```
sabellius, line 54: I would like to discuss this, do we want this work quick and dirty just for card cover? or set up the road to generally upload attachments to cards? because currently, there is no real attachment management for attachments, a user can keep uploading and overriding a cards cover while previous cover will silently disappear from the apps existence, while being kept forever in the cloud storage. A semi-decent approach would be to save uploaded files/images on an array of `card.attachments` and in the cover reference or denormalize the relevant single attachment.

Obviously the "correct" approach will probably take longer to implement, and since we got very limited time to do things, are we ok with going with this half baked solution? @snoxlax @LiorRey

frontend/src/components/AttachmentMenu.jsx
```diff
@@ -71,7 +115,10 @@ export function AttachmentMenu({
                 className={`cover-size-preview ${
                   textOverlay === true ? "selected" : ""
                 }`}
-                style={{ backgroundColor: coverColor }}
+                style={{
+                  backgroundColor: coverImg ? undefined : coverColor,
+                  backgroundImage: coverImg ? `url(${coverImg})` : undefined,
+                }}
```
sabellius: Seems a little dubious, using `undefined` in this way.. Why not use a more simple approach?
```js
coverImg
  ? { backgroundImage: `url(${coverImg})` }
  : { backgroundColor: coverColor };
```

frontend/src/components/Card.jsx
```diff
@@ -63,8 +63,26 @@ export function Card({
...
+          ? `linear-gradient(180deg, rgba(0, 0, 0, 0.5) 0%, #000000 100%), url(${coverImg})`
+          : `url(${coverImg})`,
+      }
+    : coverColor
+    ? { backgroundColor: coverColor }
+    : undefined;
+
+  const overlayText = isOverlay
+    ? { fontWeight: 500, fontSize: "16px" }
+    : undefined;
```
sabellius: Why is all of this done in the component?

frontend/src/components/Card.jsx
```diff
@@ -105,12 +120,11 @@ export function Card({
...
                   value={title}
                   onChange={e => setTitle(e.target.value)}
+                  style={overlayText}
                 />
                 <div className="card-footer">
                   <div className="card-footer-left">
-                    <RemoveRedEyeOutlined />
-                    <ChatRounded />
-                    <NotesRounded />
+                    {card.description === "<p></p>" ? <NotesRounded /> : null}
```
sabellius, line 126: Is this the text editor default value? is there any chance the markup will be different than this even by a single whitespace character or any editor default additions than will cause the condition to be false for the wrong reason?

frontend/src/components/Card.jsx
```diff
@@ -136,16 +150,12 @@ export function Card({
             <Box
               className="card-content-container"
               onClick={handleClickCard}
-              sx={
-                card.cover?.textOverlay && { backgroundColor: card.cover.color }
-              }
+              sx={card.cover?.textOverlay ? coverStyle : undefined}
```
sabellius, line 152: This is a pattern I've seen a lot in this PR, why do you prefer this over a short circuit? `sx={card.cover?.textOverlay && coverStyle}`

frontend/src/services/attachment-service.js
```diff
@@ -0,0 +1,37 @@
...
+
+  const formData = new FormData();
+  formData.append("file", file);
+  formData.append("api_key", apiKey);
+  formData.append("timestamp", String(timestamp));
+  formData.append("signature", signature);
+  formData.append("folder", folder);
+
+  const uploadUrl = `https://api.cloudinary.com/v1_1/${cloudName}/image/upload`;
+  const res = await fetch(uploadUrl, { method: "POST", body: formData });
```
sabellius, line 27: Nothing wrong with using `fetch` but that's why we have a `http-service` for.

frontend/src/services/attachment-service.js
```diff
@@ -0,0 +1,37 @@
...
+  const data = await res.json();
+
+  if (!res.ok) {
+    const msg =
+      data?.error?.message || data?.message || "Cloudinary upload failed";
+    throw new Error(msg);
+  }
```
sabellius, line 34: Does this even work? you first get the `data` and only then check if there was an error?

frontend/src/components/AttachmentMenu.jsx
```diff
@@ -10,24 +12,52 @@ export function AttachmentMenu({
...
+
+      const uploaded = await attachmentService.uploadImage(file, "card-covers");
+
+      updateCardCover(card._id, {
+        img: uploaded.secure_url,
+        color: null,
+        textOverlay: textOverlay,
+      });
+    } catch (err) {
+      setUploadError(err?.message || "Upload failed");
```
sabellius, line 56: an `err` object will always be present so  I'm not sure why safe navigation is needed here.

Regardless, if the errors you expect are from the server, they don't contain a `message` key for sure.

sabellius (changes requested):
See my comments

frontend/src/services/attachment-service.js
snoxlax: our http service is for our backend this is a post to Cloudinary

frontend/src/services/attachment-service.js
snoxlax: Yes, why wouldn’t it work? You still get data if the response isn’t ok, the data will contain the error message or some other message returned

frontend/src/components/Card.jsx
snoxlax: I will check again but yes it always has this when its empty

frontend/src/components/Card.jsx
snoxlax: Because I need to apply the correct styling based on the condition whether there is a background color, an image the overlay is enabled or not. 
Where else should this logic live?

frontend/src/services/attachment-service.js
sabellius: This is silly, their implementation should've provided a way to override the base URL.

frontend/src/components/AttachmentMenu.jsx
snoxlax: My plan was to implement this step by step: first the cover, then using an image as the cover. After that, I will move on to attachments, allowing the user to choose which attachment to use as the cover, assuming there is enough time to complete it.

frontend/src/services/attachment-service.js
sabellius: The example in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) summarizes how I would've expected this, it only work because `fetch` doesn't raise an error on 4XX-5XX statues.
Honestly its just a personal preference I guess - you can keep this as-is.

backend/src/services/upload-service.js
snoxlax: well this is what is expected to be sent
```js
const formData = new FormData();
  formData.append("file", file);
  formData.append("api_key", apiKey);
  formData.append("timestamp", String(timestamp));
  formData.append("signature", signature);
  formData.append("folder", folder);

  const uploadUrl = `https://api.cloudinary.com/v1_1/${cloudName}/image/upload`;
  const res = await fetch(uploadUrl, { method: "POST", body: formData });
```
i dont think we should have these values stored in our frontend some where

frontend/src/components/AttachmentMenu.jsx
snoxlax: It should empty the input

frontend/src/components/Card.jsx
sabellius: Well, normally we would put style related code in css files (classes) and conditionally render classes instead of filling the component files with inline styles. I mean, sure, there is the pressure of time to finish things quickly, but does this mean we should decrease code quality by cutting corners?
That way, the components is less concerned with the actual styling but rather with applying the correct classes.
```js
const coverClasses = ["cover"];

if (coverImg) {
  coverClasses.push("cover--image");
  if (isOverlay) coverClasses.push("cover--overlay");
} else if (coverColor) {
  coverClasses.push("cover--color");
}

const coverClassName = coverClasses.join(" ");
```

Since using urls will require use of css variable and is a bit more complex, you probably should keep it like this.

frontend/src/components/AttachmentMenu.jsx
sabellius: Does this means it also resets `ev.target.files`?

frontend/src/components/AttachmentMenu.jsx
snoxlax: > If you want to avoid ever producing style={{ backgroundColor: null }}, you need to only create the style object when you have a real value, otherwise pass undefined/null for the whole style prop.
> For your case:
> 
```js 
const coverStyle = coverImg  
? { backgroundImage: `url(${coverImg})` }  
: coverColor  ? { backgroundColor: coverColor }  
: undefined; 
```
> This guarantees you won’t end up with { backgroundColor: null }.  

---AI

snoxlax: I checked with ai and it said this

backend/src/controllers/cloudinary-controller.js
snoxlax: I added another check in the service, but generally, if it reaches this error, it is a worst-case scenario in my opinion

frontend/src/components/AttachmentMenu.jsx
snoxlax: That line resets the file input after an upload:
* ev.target.value = "" clears the <input type="file">’s selected file, so the control becomes “empty” again.

It’s not strictly required for the upload to work, but it’s useful UX/behavior because without it:
* If the user selects the same file again, many browsers won’t fire onChange (because the value didn’t change), so your handleUploadCover won’t run.

So: keep it unless you intentionally want to prevent re-selecting the same file without picking a different one first.

backend/src/controllers/upload-controller.js
```diff
@@ -0,0 +1,15 @@
...
+
+const ALLOWED_FOLDERS = new Set(["card-covers"]);
+
+export async function getUploadSignature(req, res) {
+  const folder = req.body?.folder;
+  if (!folder) throw createError(400, "folder is required");
+  if (!ALLOWED_FOLDERS.has(folder)) throw createError(400, "Invalid folder");
+
+  const signData = uploadService.generateUploadSignature(folder);
+  if (!signData) throw createError(500, "Failed to generate upload signature");
```
sabellius, line 12: It looks like an unreachable piece of code. `uploadService.generateUploadSignature` is always returning an object (maybe with some `undefined` attributes, but returns it nonetheless). If an error occurs in that function, you won't get to this condition either, as the error will bubble up through here to the error-handler middleware. It's not a bug, it just doesn't do anything unless I'm missing something.

sabellius (approved):
Lookd good, see my last comment.

PR #98, opened 2025-12-17
Applied fix for description icon

snoxlax:
## 🎯 Description
This is a fix for the icon in description to show it when there is a description and hide when there isnt one


LiorRey (approved):
After reviewing, I approve 🦾

PR #101, opened 2025-12-18
feature/CATC_151-implement-filters-in-the-backend

sabellius:
## 🎯 Description
Implement card filtering in the backend and fix critical issues in the frontend.

## 🔧 Changes
- Imlement a filter service and functionality for filtering card inside `getFullById`
- **🚫 Exclude current user from member list**: Current user is no longer shown in the member autocomplete dropdown since there's already a separate "Assigned to me" checkbox
- **✅ Fix select all checkbox behavior**: The select all checkbox now properly toggles between selecting all available members and clearing the selection, while preserving the "Assigned to me" state when appropriate
- **🎯 Improve checkbox icon states**: Fixed the indeterminate state calculation to correctly show when some (but not all) members are selected
- **🧹 Remove redundant filtering**: Eliminated unnecessary filtering operations for better performance

## 📝 Notes
The changes ensure that:
- Users won't see themselves duplicated in both the member list and "Assigned to me" option
- The select all checkbox provides clear visual feedback (checked/unchecked/indeterminate)
- When using the select all checkbox, it respects the existing "Assigned to me" selection

## 🖥️ Demo

https://github.com/user-attachments/assets/e207a3a6-71d3-4d84-9385-08a5b066fbf7

### DIFF
--- a/backend/src/controllers/board-controller.js
+++ b/backend/src/controllers/board-controller.js
@@ -45,7 +45,30 @@ export async function getBoardById(req, res) {
 }
 
 export async function getFullBoardById(req, res) {
-  const fullBoard = await boardService.getFullBoardById(req.params.id);
+  const { title, labels, members, noMembers, includeNoLabels } = req.query;
+
+  const filterBy = {
+    title,
+    labels: labels
+      ? labels
+          .split(",")
+          .map(id => id.trim())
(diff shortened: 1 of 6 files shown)


snoxlax (approved):
Looks good! only remove the comments where not needed

PR #102, opened 2025-12-18
feature/CATC_152-build-frontend-directly-into-the-backend-public-folder-for-serving

sabellius:
## 🎯 Description
This PR makes preparations to enable deploying the app to Render.

---

## 🔧 Changes
- Move vite from dev dependencies to regular dependencies as it will be used during deployment.
- Change vite build config to put the build artifacts directly in the backends' `public/` folder.
- Remove data generators and imports of local services as they use `faker.js` which is a dev dependency, and causes an error during the deployment as it's not installed. (We also agreed that we dont longer need those services altogether).
- Serve the React build files (index.html) using a wild card route from the backend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ dist/
 dist-ssr/
 build/
 *.tsbuildinfo
+public/
 
 # Vite (Frontend)
 .vite/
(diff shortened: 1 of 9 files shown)


snoxlax (approved):
All good looks good!

PR #110, opened 2025-12-20
bugfix/CATC_157-editing-card-title-causes-other-elements-to-shift

sabellius:
## 🎯 Description
Fixes UI bug where editing card titles causes other modal elements to shift.

## 🔧 Changes
- **🎨 CSS Update**: Removed borders and background from title input in normal state, making it appear as plain text while maintaining consistent element rendering
- **⚛️ Component Refactor**: Eliminated conditional rendering between `<h1>` and `<TextareaAutosize>` - now always renders textarea to prevent layout shifts without setting fixed heights.
- **🔄 State Simplification**: Removed `isEditing` state management, simplifying the title editing flow

## 📝 Notes
The root cause was switching between different HTML element types (h1 vs textarea) which triggered layout recalculation. By using a single element type and controlling appearance via CSS, the layout remains stable during title editing.
Description was not affected by this PR as other teammates are making modifications that might affect any attempts at fixed it moving modal elements as well.

## 🖥️ Demo

https://github.com/user-attachments/assets/24cc64ac-92df-452c-a7b6-ab0685acbdaf

### DIFF
--- a/frontend/src/assets/styles/components/CardModal.css
+++ b/frontend/src/assets/styles/components/CardModal.css
@@ -186,8 +186,8 @@
   word-wrap: break-word;
   word-break: break-word;
   color: var(--gray1);
-  background-color: var(--gray5, #040a10);
-  border: 1px solid var(--form-input-border-color, #4a9eff);
+  border: none;
+  outline: none;
   border-radius: 4px;
   resize: none;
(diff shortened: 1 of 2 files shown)


snoxlax (approved):
Looks good!

PR #113, opened 2025-12-21
Add Comments in card modal

snoxlax:
## 🎯 Description
Add comment functionality to the card modal, allowing authenticated users to manage comments on individual cards.

## 🔧 Changes
* Added the ability to **add, edit, and delete comments** within the card modal.
* Each comment is associated with the **currently connected (authenticated) user**.
* Only the comment owner can see and use the **edit** and **delete** actions on their own comments.
* Introduced a **dedicated Comment Section component** with its own styling.
* Reused the existing **rich text editor** from the description section:

  * Adjusted sizing and layout to better fit comment usage.
  * The same editor is used for both **adding** and **editing** comments.
 
## 📝 Notes
* Comment permissions are enforced at the UI level to ensure users can only modify their own comments.
* The implementation follows existing editor patterns to maintain consistency across the card modal.

## 🖥️ Demo
<details>

  <p align="center">
  </p>

</details>

### DIFF
--- a/backend/src/models/Card.js
+++ b/backend/src/models/Card.js
@@ -30,13 +30,6 @@ const commentSchema = new mongoose.Schema(
   { timestamps: true }
 );
 
-commentSchema.pre("save", function (next) {
-  if (!this.isNew && this.isModified("text")) {
-    this.isEdited = true;
-  }
-  next();
-});
-
 const attachmentSchema = new mongoose.Schema(
   {
     url: {
--- a/backend/src/services/card-service.js
+++ b/backend/src/services/card-service.js
@@ -107,7 +107,7 @@ export async function addComment(cardId, user, text) {
   card.comments.push(newComment);
   await card.save();
 
-  return card.comments[card.comments.length - 1];
+  return card;
 }
 
 export async function updateComment(cardId, commentId, text) {
@@ -120,7 +120,7 @@ export async function updateComment(cardId, commentId, text) {
   comment.text = text.trim();
   await card.save();
 
-  return comment;
+  return card;
 }
 
 export async function deleteComment(cardId, commentId) {
@@ -133,7 +133,7 @@ export async function deleteComment(cardId, commentId) {
   card.comments.pull(commentId);
   await card.save();
 
-  return true;
+  return card;
 }
 
 export async function getComments(cardId) {
--- a/backend/src/services/card-service.js
+++ b/backend/src/services/card-service.js
@@ -117,7 +117,11 @@ export async function updateComment(cardId, commentId, text) {
   const comment = card.comments.id(commentId);
   if (!comment) throw createError(404, "Comment not found");
 
-  comment.text = text.trim();
+  const trimmedText = text.trim();
+  if (comment.text !== trimmedText) {
+    comment.text = trimmedText;
+    comment.isEdited = true;
+  }
   await card.save();
 
   return card;
--- a/frontend/src/components/CardComments.jsx
+++ b/frontend/src/components/CardComments.jsx
@@ -1,0 +1,211 @@
+import { useState } from "react";
+import Button from "@mui/material/Button";
+import { useSelector } from "react-redux";
+import { TextEditor } from "./ui/TextEditor";
+import { Avatar } from "./ui/Avatar";
+import {
+  addComment,
+  updateComment,
+  deleteComment,
+} from "../store/actions/board-actions";
+import { selectCurrentUser } from "../store/selectors/auth-selectors";
+
+export function CardComments({ card }) {
+  const currentUser = useSelector(selectCurrentUser);
+  const [newCommentText, setNewCommentText] = useState(null);
+  const [isAddingComment, setIsAddingComment] = useState(false);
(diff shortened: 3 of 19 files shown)


backend/src/models/Card.js
```diff
@@ -30,13 +30,6 @@ const commentSchema = new mongoose.Schema(
   { timestamps: true }
 );
 
-commentSchema.pre("save", function (next) {
-  if (!this.isNew && this.isModified("text")) {
-    this.isEdited = true;
-  }
-  next();
-});
-
```
snoxlax, line 39: This didnt work well it made all other comments edited when we add a new comment

backend/src/services/card-service.js
```diff
@@ -117,10 +117,14 @@ export async function updateComment(cardId, commentId, text) {
   const comment = card.comments.id(commentId);
   if (!comment) throw createError(404, "Comment not found");
 
-  comment.text = text.trim();
+  const trimmedText = text.trim();
+  if (comment.text !== trimmedText) {
+    comment.text = trimmedText;
+    comment.isEdited = true;
+  }
```
snoxlax, line 124: This is instead of the pre("save")

frontend/src/components/CardPopover.jsx
```diff
@@ -161,7 +161,7 @@ export function CardPopover({
           horizontal: "left",
         }}
         sx={{
-          zIndex: theme => theme.zIndex.modal + 2,
+          zIndex: theme => theme.zIndex.modal + 1,
```
snoxlax, line 164: @sabellius Why did you change this to 2? this broke something and it made me confused
it works now

backend/src/models/Card.js
sabellius: Can you elaborate, on how this happened?

frontend/src/components/CardComments.jsx
```diff
@@ -0,0 +1,211 @@
...
+  addComment,
+  updateComment,
+  deleteComment,
+} from "../store/actions/board-actions";
+import { selectCurrentUser } from "../store/selectors/auth-selectors";
+
+export function CardComments({ card }) {
+  const currentUser = useSelector(selectCurrentUser);
+  const [newCommentText, setNewCommentText] = useState(null);
+  const [isAddingComment, setIsAddingComment] = useState(false);
```
sabellius, line 16: Generally speaking, that's why we have a loading state in the global state.

frontend/src/components/CardComments.jsx
```diff
@@ -0,0 +1,211 @@
...
+  const [isCommentEditorOpen, setIsCommentEditorOpen] = useState(false);
+  const [editingCommentId, setEditingCommentId] = useState(null);
+  const [editingCommentText, setEditingCommentText] = useState(null);
```
sabellius, line 19: Does this work correctly with a `null`? Shouldn't this be a string?

frontend/src/components/CardComments.jsx
```diff
@@ -0,0 +1,211 @@
...
+  const [isSavingEdit, setIsSavingEdit] = useState(false);
+
+  async function handleAddComment() {
+    if (!newCommentText || isAddingComment) return;
```
sabellius, line 23: Is there anything that prevent the user from submitting a string full of spaces/whitespace characters?

frontend/src/components/CardComments.jsx
```diff
@@ -0,0 +1,211 @@
...
+    if (!currentUser || !comment.author) return false;
+    return (
+      comment.author.userId === currentUser._id ||
+      comment.author.userId?.toString() === currentUser._id?.toString()
+    );
+  }
+
+  function formatCommentDate(dateInput) {
+    const date = dateInput ? new Date(dateInput) : null;
+    if (!date || Number.isNaN(date.getTime())) return "";
```
sabellius, line 83: Why the second part is needed? what happens if line 82 doesn't happen correctly?

frontend/src/components/CardPopover.jsx
sabellius: There is some issue with the current z-index, you can see the scroll bar of the list on top of it:

Image

frontend/src/store/reducers/board-reducer.js
```diff
@@ -375,6 +378,66 @@ const handlers = {
...
+      ...state.board,
+      lists: state.board.lists.map(list =>
+        list._id === action.payload.listId
+          ? {
+              ...list,
+              cards: list.cards.map(card =>
+                card._id === action.payload._id
+                  ? { ...card, comments: action.payload.comments }
+                  : card
+              ),
```
sabellius, line 435: Since you decided to return an entire card from the server and the remote-service, There is no need to overwrite the comments specifically, you can just return the entire payload for that specific card id.
This is probably relevant to the other reducers as well.

sabellius (changes requested):
Please see my comments.

LiorRey (approved):
After reviewing the code and apart from @sabellius comments (which I mostly agree with them) I approve 🦾

sabellius (approved):

